PR #327, opened 2026-03-10
fix: ensure pytest-jira processes dynamically-added markers

krcmarik:
Add pytest_collection_modifyitems(tryfirst=True) hook that applies
jira markers from config to collected test items. tryfirst=True is
required because pytest-jira's own collection hook must find the
markers already present on items; without it, pytest-jira runs first,
finds no jira markers, and never processes them.

## Summary by CodeRabbit

* **Chores**
  * Adjusted test collection ordering so provider-based test skipping runs earlier during collection, improving test reliability and reducing spurious executions across environments. This change optimizes test discovery timing without altering test behavior or public test interfaces.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -255,6 +255,7 @@ def pytest_sessionfinish(session, exitstatus):
                 LOGGER.exception("Failed to enrich JUnit XML, original preserved")
 
 
+@pytest.hookimpl(tryfirst=True)
 def pytest_collection_modifyitems(session, config, items):
     # -------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


krcmarik:
> We need to do branches for versions, which will be proper solution for this and many other things.

@myakove True, I am just trying to address the change in [MTV-2846](https://issues.redhat.com/browse/MTV-2846) which will may be soon closes and fixed in 2.12 but not fixed in 2.11, 2.10 which will cause warm migration tests executed for 2.10/2.11. Or do we want to move to branches per version relatively soon?

PR #336, opened 2026-03-12
fix: make labeled_worker_node fixture parallel-safe

krcmarik:
Append session UUID suffix to the label key when in auto-generated mode (config_value is None), so two parallel workers that select the same node no longer overwrite each other's label. Truncation now preserves the UUID suffix (the unique part) by shortening the base key instead.

## Summary by CodeRabbit

* **Refactor**
  * Worker node labeling now auto-generates consistent, session-based labels when unset; qualified keys preserve their prefix while ensuring name length limits.
* **Bug Fixes**
  * Added validation to require exactly one target label, preventing ambiguous node selection and raising clear errors for multiple/missing labels.
* **Documentation**
  * Updated guidance: default label values may be auto-generated and are trimmed as needed to fit constraints; test-id removal noted.

### DIFF
--- a/README.md
+++ b/README.md
@@ -482,7 +482,7 @@ Schedule VMs to specific labeled nodes.
 
 ```python
 "target_node_selector": {
-    "migration-workload": None,  # Auto-generated with session_uuid
+    "migration-workload": None,  # Exactly one label; None = auto-generated key & value with session_uuid
 }
 ```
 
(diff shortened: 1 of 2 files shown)


redhat-qe-bot1:
Cherry-picked PR fix: make labeled_worker_node fixture parallel-safe into v2.11: https://github.com/RedHatQE/mtv-api-tests/pull/337

PR #499, opened 2026-05-18
refactor: remove explicit access mode from copy-offload storage maps

tshefi:
Forklift defaults to ReadWriteMany when accessMode is not set in the StorageMap destination, so the explicit access_mode="ReadWriteOnce" was both unnecessary and incorrect.

Also fix post_migration.py check to assert RWO only when accessMode is explicitly set to "ReadWriteOnce", not for any non-empty accessMode value.


tshefi:
Test

Issue #306, opened 2026-02-25
feat: apply KubeletConfig for image GC thresholds in autouse fixture


myakove:
## Summary

Create a session-scoped autouse fixture that applies a `KubeletConfig` resource to lower the kubelet's image garbage collection thresholds, preventing must-gather disk usage failures.

## Problem / Motivation

Must-gather fails with disk usage errors because the kubelet's default `imageGCHighThresholdPercent` (85%) is higher than must-gather's built-in disk checker threshold (70%). Nodes accumulate stale container images (especially `redhat-operator-index` from OLM CatalogSource updates) and sit at ~80-84% disk usage — below kubelet GC but above must-gather's limit.

Error observed:
```
[disk usage checker] Volume usage percentage: current = 83 ; allowed = 70
[disk usage checker] Disk usage exceeds the volume percentage of 70 for mounted directory, terminating...
```

## Requirements

1. Create a `KubeletConfig` resource that sets:
   - `imageGCHighThresholdPercent: 70`
   - `imageGCLowThresholdPercent: 60`
2. Apply this configuration via an autouse session-scoped fixture so it's set at the start of every test run
3. Use `create_and_store_resource()` for proper resource tracking and teardown
4. The fixture should be idempotent — if the `KubeletConfig` already exists with the correct values, skip creation

## Deliverables

- [ ] Add `KubeletConfig` resource creation utility or fixture
- [ ] Add autouse session-scoped fixture that applies the GC thresholds
- [ ] Ensure the fixture is idempotent (no-op if config already matches)
- [ ] Update CLAUDE.md if new patterns are introduced

## Notes

- The `KubeletConfig` is a cluster-level resource in OpenShift that modifies kubelet parameters on nodes matching a `MachineConfigPool` selector
- Lowering the GC thresholds ensures kubelet proactively cleans up unused images before disk usage reaches must-gather's 70% limit
- The 60%/70% low/high thresholds provide a 10% hysteresis band to avoid thrashing

PR #536, opened 2026-06-01
fix: limit NIC name detection to preserve_static_ips tests

MiriSafra:
## Summary
- Gate `detect_guest_nic_names` (pre-migration) and `check_nic_name_preservation` (post-migration) on `preserve_static_ips` config
- NIC name preservation udev rules are generated by the same firstboot script as static IP preservation, so they share the same precondition
- Fixes regression from #524 where `detect_guest_nic_names` crashed setup on clone-based tests (e.g. copy-offload) where VMware Tools hadn't reported `guestFamily` yet

## Bug Fixes
- Updated network interface preservation validation logic to properly align with static IP preservation configuration.
- Refined conditional checks for guest operations IP detection during migration, ensuring operations only execute when static IP preservation is enabled.

## Refactor
- Consolidated conditional logic for network operations to follow consistent static IP preservation checks.


conftest.py
```diff
@@ -1150,12 +1150,13 @@ def prepared_plan(
                         "Static IP verification may not work for this VM."
                     )
 
-                detect_guest_nic_names(
-                    source_provider=source_provider,
-                    vm=provider_vm_api,
-                    source_provider_data=fixture_store["source_provider_data"],
-                    vm_details=source_vm_details,
-                )
+                if plan.get("preserve_static_ips"):
```
krcmarik, line 1153: The gating on preserve_static_ips makes sense - NIC name detection and static IP origin detection share the same precondition of a running source guest with tools active. I believe tho that It's currently an undocumented assumption - preserve_static_ips implicitly requires source_vm_power: "on" in the test config - without a running guest, neither static IP origins nor NIC names can be collected. We should consider making it explicit with a validation in the prepared_plan fixture (validating that if we set preserve_static_ips in the text config source_vm_power is set to "on" too) or alternatively, document this requirement (or both).

tshefi:
/lgtm 

Offload test pass using your fork, waiting for merge.
^ @prabinovRedhat  job  #306 migration plan reached running state.

conftest.py
MiriSafra: Makes sense. I opened a separate PR- #537 for the validation to keep this one focused on the fix

redhat-qe-bot1:
Test Oracle server is not responding, skipping test analysis

redhat-qe-bot1:

prabinovRedhat:
please check if you need cherry pick this fix @MiriSafra

MiriSafra:
> please check if you need cherry pick this fix @MiriSafra

@prabinovRedhat thanks, no cherry-pick needed, #524 was only on main

PR #550, opened 2026-06-09
feat(MTV-5654): add copy-offload populator throttling test

prabinovRedhat:
## Summary

Adds MTV-5654 copy-offload **(Case 1 + Case 3)** coverage for per-ESXi-host populator throttling and `sourceHost` labels.

- **`TestCopyoffloadPopulatorThrottlingMigration`**: 7-step flow (storagemap → networkmap → plan → migrate with peak monitoring → verify throttling → check xcopy → check VMs)
- **`populator_inflight_forkliftcontroller` fixture**: class-scoped patch of `ForkliftController.spec.controller_max_populator_inflight` and deployment env to limit **2**, with pre-test snapshot restore on teardown and xdist file lock
- **Verification utilities**: peak in-flight populate pods during migration, `PopulatorThrottled` events (≥ disks − limit), single `sourceHost` label on all populate pods
- **`wait_for_migration_complate(on_status_poll=...)`** and public **`get_migration_for_plan()`** for migration-scoped monitoring

Rebased on latest `main` (includes `merge_storage_secret_extra` integration).

## Test plan

- [x] Local cluster run: `uv run pytest -m copyoffload -k TestCopyoffloadPopulatorThrottlingMigration` — **7/7 passed** (~17 min)
- [x] Peak concurrency observed: **2/2** with 5 disks (1 boot + 4 add_disks)
- [x] **3** `PopulatorThrottled` events with `(2/2 in-flight)`
- [x] Fixture restore: CR/deployment limit **2 → 20** after class teardown

## Notes

- **Cluster-mutating test**: temporarily sets global populator in-flight limit to 2. Do not run in parallel with other copy-offload tests on the same cluster.
- MTV-5654 Test 2/4 (multi-host) are out of scope for this PR.

## Summary by CodeRabbit

* **New Features**
  * Added end-to-end coverage for copy-offload populator throttling, including per-host inflight concurrency validation and expected throttling/event behavior.
  * Introduced shared testing utilities to temporarily control and verify populator inflight limits safely during throttling scenarios.
* **Tests**
  * Added a new populator-throttling test scenario and multi-disk VM setup for copy-offload migrations.
  * Improved concurrent migration flows to wait for copy-offload plan populator secrets before validation.
* **Documentation**
  * Updated contributor/testing guidance with a step-by-step throttling test pattern and added a copy-offload warning about cluster-wide throttle mutations.


tests/copyoffload/test_copyoffload_migration.py
```diff
@@ -3472,6 +3478,203 @@ def test_check_vms(
...
+            destination_provider=destination_provider,
+            ocp_admin_client=ocp_admin_client,
+            source_provider_inventory=source_provider_inventory,
+            vms=vms_names,
+            storage_class=storage_class,
+            datastore_id=datastore_id,
+            offload_plugin_config=offload_plugin_config,
+            volume_mode="Block",
+        )
+        assert self.storage_map, "StorageMap creation failed"
```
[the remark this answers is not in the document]
prabinovRedhat: Good catch - agreed. The boilerplate steps were copied from older copy-offload classes that predate our typing convention; only the new populator-specific methods were annotated initially.

Fixed in b9c4dc0: all methods in TestCopyoffloadPopulatorThrottlingMigration now have complete fixture parameter types and -> None, aligned with newer classes like TestCopyoffloadRdmPhysicalDiskMigration.

guides/copyoffload/how-to-run-copyoffload-tests.md
```diff
@@ -402,10 +402,35 @@ The Job automatically reads cluster credentials from the Secret created in Step
...
+
+**Do not** run this test in parallel with other tests that depend on the default populator limit on the
+same cluster. Run it explicitly:
+
+```bash
+uv run pytest -m copyoffload -k TestCopyoffloadPopulatorThrottlingMigration -v \
+  --providers-json .providers.json \
+  --tc=source_provider:<your-vsphere-provider-key> \
+  --tc=storage_class:<your-block-storage-class>
+```
```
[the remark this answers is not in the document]
prabinovRedhat: outdated, there is no code like this anymore

prabinovRedhat:
cc: @tshefi

tests/copyoffload/conftest.py
```diff
@@ -148,6 +160,262 @@ def multi_datastore_config(source_provider_data: dict[str, Any]) -> None:
...
+
+def _forkliftcontroller_populator_inflight_lock_path() -> Path:
+    """Return the cross-worker lock path for ForkliftController populator limit changes."""
+    lock_dir = Path(tempfile.gettempdir()) / "pytest-shared-forklift"
+    _ensure_secure_shared_lock_dir(lock_dir=lock_dir)
+    return lock_dir / "populator-inflight.lock"
+
+
+_POPULATOR_CONTROLLER_DEPLOYMENT = "forklift-volume-populator-controller"
+_MAX_POPULATOR_INFLIGHT_ENV = "MAX_POPULATOR_INFLIGHT"
```
myakove: Move the module-level constants (`_POPULATOR_CONTROLLER_DEPLOYMENT`, `_MAX_POPULATOR_INFLIGHT_ENV`) to the top of the file, right after the imports section.

tests/copyoffload/conftest.py
```diff
@@ -148,6 +160,262 @@ def multi_datastore_config(source_provider_data: dict[str, Any]) -> None:
     LOGGER.info("✓ Multi-datastore configuration validated: secondary_datastore_id = %s", secondary_datastore_id)
 
 
+def _ensure_secure_shared_lock_dir(lock_dir: Path) -> None:
```
myakove: `conftest.py` should contain only pytest fixtures and hooks. Please extract all private helper functions (`_ensure_secure_shared_lock_dir`, `_forkliftcontroller_populator_inflight_lock_path`, `_get_populator_inflight_from_deployment`, `_wait_for_populator_inflight_deployment`, `_restore_forkliftcontroller_populator_inflight`) to a utility module (e.g., `utilities/`) and import them here.

tests/copyoffload/conftest.py
```diff
@@ -148,6 +160,262 @@ def multi_datastore_config(source_provider_data: dict[str, Any]) -> None:
...
+    if current_limit == restore_cr_limit:
+        LOGGER.info(
+            f"ForkliftController controller_max_populator_inflight already {restore_cr_limit!r} (pre-test value)"
+        )
+    else:
+        LOGGER.info(
+            f"Restoring ForkliftController controller_max_populator_inflight "
+            f"from {current_limit!r} to pre-test value {restore_cr_limit!r}"
+        )
+        ResourceEditor(
```
myakove: `ResourceEditor` supports automatic restore via context manager (`with ResourceEditor(...) as editor:`). Consider using it instead of calling `.update(backup_resources=False)` with manual restore logic in `_restore_forkliftcontroller_populator_inflight` — the `with` block handles rollback automatically on exit.

utilities/copyoffload_migration.py
```diff
@@ -691,3 +767,274 @@ def verify_xcopy_used_per_datastore(
             "Migration must include at least one disk from each configured datastore; "
             f"verified datastore IDs: {sorted(verified_datastore_ids)}"
         )
+
+
+_ACTIVE_POPULATOR_POD_PHASES = frozenset({"Running", "Pending"})
```
myakove: Move `_ACTIVE_POPULATOR_POD_PHASES` to the top of the module with the other constants, rather than defining it inline before its first usage at line 770.

myakove (changes requested):

tests/copyoffload/conftest.py
```diff
@@ -148,6 +160,262 @@ def multi_datastore_config(source_provider_data: dict[str, Any]) -> None:
...
+) -> None:
+    """Wait until the populator controller deployment applies the in-flight limit.
+
+    ForkliftController spec changes propagate to the populator-controller Deployment
+    asynchronously. Migration must not start until MAX_POPULATOR_INFLIGHT is active.
+
+    Args:
+        ocp_admin_client (DynamicClient): OpenShift admin client.
+        mtv_namespace (str): Namespace where the populator controller runs.
+        expected_limit (int): Expected MAX_POPULATOR_INFLIGHT value.
```
myakove: **Operator precedence bug:** This evaluates as `(deployment_status.availableReplicas or 0) if deployment_status else 0` due to Python precedence. If `deployment_status` exists but `availableReplicas` is `None` (e.g., during rollout), it works by accident — but the intent is ambiguous.

Use explicit parentheses or split:
```suggestion
        available_replicas = (deployment_status.availableReplicas or 0) if deployment_status else 0
```
Or clearer:
```python
available_replicas = 0
if deployment_status:
    available_replicas = deployment_status.availableReplicas or 0
```

utilities/mtv_migration.py
```diff
@@ -260,6 +275,8 @@ def create_plan_resource(
 
     # Wait for Forklift to create plan-specific secret for copy-offload (race condition)
     if copyoffload:
+        from utilities.copyoffload_migration import wait_for_plan_secret  # noqa: PLC0415
```
myakove: Inline import — all imports should be at the top of the file. If this is a circular import workaround, consider moving `wait_for_plan_secret` to a module that doesn't import from `mtv_migration`, or restructure to break the cycle.

myakove (commented):
## Code Review

### :bulb: Suggestions (2)

| File | Line | Issue |
|------|------|-------|
| `tests/copyoffload/conftest.py` | 230 | **Operator precedence bug:** This evaluates as `(deployment_status.availableRepl |
| `utilities/mtv_migration.py` | 278 | Inline import — all imports should be at the top of the file. If this is a circu |

---
*Review generated by pi*

utilities/mtv_migration.py
prabinovRedhat: Agreed — this was a circular import workaround. Moved wait_for_plan_secret to utilities/copyoffload_plan_secret.py so mtv_migration can import it at module level without pulling in copyoffload_migration.

tests/copyoffload/conftest.py
prabinovRedhat: Good catch — replaced the ambiguous ternary/or expression with an explicit if deployment_status block in wait_for_populator_inflight_deployment().

utilities/copyoffload_migration.py
prabinovRedhat: Done — moved _ACTIVE_POPULATOR_POD_PHASES to the module constants section at the top.

tests/copyoffload/conftest.py
prabinovRedhat: Good point — switched to with ResourceEditor(...) via a populator_inflight_limit() context manager. The CR now rolls back automatically on exit; we still wait for the populator deployment to reconcile after restore since that is separate from the CR patch.

tests/copyoffload/conftest.py
prabinovRedhat: Agreed — conftest should stay fixtures-only per project guidelines. Extracted the populator ForkliftController helpers to utilities/forklift_controller_populator.py and updated the fixture to import from there.

tests/copyoffload/conftest.py
prabinovRedhat: Done — moved _POPULATOR_CONTROLLER_DEPLOYMENT and _MAX_POPULATOR_INFLIGHT_ENV to the top of the file, after the imports / LOGGER definition

utilities/copyoffload_plan_secret.py
```diff
@@ -0,0 +1,44 @@
...
+    LOGGER.info("Copy-offload: waiting for Forklift to create plan-specific secret...")
+    try:
+        for _ in TimeoutSampler(
+            wait_timeout=60,
+            sleep=2,
+            func=lambda: any(
+                s.name.startswith(f"{plan_name}-") for s in Secret.get(client=ocp_admin_client, namespace=namespace)
+            ),
+        ):
+            break
```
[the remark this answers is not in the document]
prabinovRedhat: Good catch — the unconditional break was a regression from extracting wait_for_plan_secret() into copyoffload_plan_secret.py during the review refactor. It caused the loop to exit after the first sample regardless of the result, so the 60s poll never ran when the secret wasn't ready yet.

Fixed by only returning when sample is truthy, so TimeoutSampler keeps retrying every 2s until the plan secret appears or the timeout is hit (still logs a warning and continues on timeout, as documented).

redhat-qe-bot1:
**Clean rebase detected** — no code changes compared to previous head (`4ea782c`).
The following labels were preserved: `commented-coderabbitai[bot]`.

utilities/forklift_controller_populator.py
```diff
@@ -0,0 +1,195 @@
...
+MAX_POPULATOR_INFLIGHT_ENV = "MAX_POPULATOR_INFLIGHT"
+
+
+def ensure_secure_shared_lock_dir(lock_dir: Path) -> None:
+    """Validate permissions on a cross-worker shared lock directory.
+
+    Args:
+        lock_dir (Path): Directory used for pytest-xdist file locks.
+
+    Raises:
```
myakove, line 64: TOCTOU race: the `mkdir` → `is_symlink` → `lstat` → `chmod` sequence has time-of-check-to-time-of-use gaps. An attacker could replace the directory with a symlink between `mkdir()` and `is_symlink()`. Since this is test infrastructure in controlled CI, risk is minimal — but consider adding a comment noting the limitation, or using `os.open()` with `O_NOFOLLOW` + `os.fstat()` on the fd for an atomic check.

myakove (commented):
## Code Review

### :bulb: Suggestions (2)

| File | Line | Issue |
|------|------|-------|
| `tests/copyoffload/test_copyoffload_migration.py` | 3484 | **Syntax bug:** `py_config["tests_params","test_copyoffload_populator_throttling |
| `utilities/forklift_controller_populator.py` | 33 | TOCTOU race: the `mkdir` → `is_symlink` → `lstat` → `chmod` sequence has time-of |

---
*Review generated by pi*

utilities/copyoffload_migration.py
```diff
@@ -377,19 +359,82 @@ def _get_migration_uid(plan: Plan) -> str:
...
+
+    Args:
+        plan (Plan): The Plan CR resource.
+
+    Returns:
+        str | None: Migration UID when found, otherwise None.
+    """
+    try:
+        return _get_migration_uid(plan=plan)
+    except ValueError:
```
myakove: `_resolve_migration_uid` silently swallows `ValueError` with a bare `pass` — the original error context is lost entirely. Consider logging at DEBUG level so operators can trace why the primary UID resolution failed:

```suggestion
    except ValueError as exc:
        LOGGER.debug("Primary migration UID lookup failed: %s; falling back to Migration CR", exc)
```

utilities/copyoffload_migration.py
```diff
@@ -691,3 +736,271 @@ def verify_xcopy_used_per_datastore(
...
+        namespace=target_namespace,
+        plan_name=plan.name,
+        plan_namespace=plan.namespace,
+        cut_over=cut_over,
+    )
+
+    max_concurrent_by_host: dict[str, int] = defaultdict(int)
+    migration_uid: str | None = None
+
+    def _poll_populator_concurrency(_status: str) -> None:
```
myakove: The `_poll_populator_concurrency` inner function mutates outer state via `nonlocal migration_uid` and the `max_concurrent_by_host` defaultdict. This increases cognitive complexity and makes the polling logic harder to test in isolation.

Consider extracting a small stateful class (e.g., `_PopulatorConcurrencyTracker`) with an explicit `poll(status)` method and a `results` property. That would be easier to unit-test and reason about, and avoids the `nonlocal` pattern.

utilities/forklift_controller_populator.py
```diff
@@ -0,0 +1,195 @@
...
+
+    Args:
+        forklift_controller (ForkliftController): ForkliftController resource to patch.
+        ocp_admin_client (DynamicClient): OpenShift admin client.
+        mtv_namespace (str): Namespace where the populator controller runs.
+        test_limit (int): Limit to apply for the test (e.g. POPULATOR_INFLIGHT_LIMIT).
+        original_deployment_limit (int): MAX_POPULATOR_INFLIGHT value before the test.
+    """
+    current_cr_limit = getattr(forklift_controller.instance.spec, "controller_max_populator_inflight", None)
+
```
myakove, line 302: Potential type mismatch: `current_cr_limit` comes from `getattr(forklift_controller.instance.spec, ...)` which may return a string (e.g. `"2"`) depending on how openshift-python-wrapper deserializes the Kubernetes API value. `test_limit` is always `int`. In Python `"2" == 2` is `False`, so the "already set" optimization would silently never trigger and the CR gets re-patched every time.

Consider normalizing:
```python
if current_cr_limit is not None and int(current_cr_limit) == test_limit:
```

myakove (commented):
## Code Review

### :bulb: Suggestions (3)

| File | Line | Issue |
|------|------|-------|
| `utilities/copyoffload_migration.py` | 376 | `_resolve_migration_uid` silently swallows `ValueError` with a bare `pass` — the |
| `utilities/copyoffload_migration.py` | 813 | The `_poll_populator_concurrency` inner function mutates outer state via `nonloc |
| `utilities/forklift_controller_populator.py` | 164 | Potential type mismatch: `current_cr_limit` comes from `getattr(forklift_control |

---
*Review generated by pi*

tests/copyoffload/test_copyoffload_migration.py
```diff
@@ -3472,6 +3478,203 @@ def test_check_vms(
         )
 
 
+@pytest.mark.vsphere
+@pytest.mark.copyoffload
+@pytest.mark.incremental
+@pytest.mark.parametrize(
```
prabinovRedhat, line 3485: Thanks for catching this — agreed that a tuple key py_config["tests_params", "..."] would break collection.

In the current code we already use nested access: py_config["tests_params"]["test_copyoffload_populator_throttling_migration"] (see line 3486), consistent with the other copy-offload test classes. I verified this in every commit on the branch and in a full manual run (7/7 collected and passed).

prabinovRedhat: Thanks for catching this — agreed that a tuple key py_config["tests_params", "..."] would break collection.

In the current code we already use nested access: py_config["tests_params"]["test_copyoffload_populator_throttling_migration"] (see line 3486), consistent with the other copy-offload test classes. I verified this in every commit on the branch and in a full manual run (7/7 collected and passed).

utilities/forklift_controller_populator.py
prabinovRedhat: Good point on the TOCTOU gap between mkdir() and path-based checks.

Updated to open the directory with O_NOFOLLOW | O_DIRECTORY, validate ownership via fstat() on the fd, and set mode with fchmod() on that same fd — so we don't act on a path that was swapped to a symlink after creation. Documented in the docstring that a small mkdir→open window remains, which we accept for this pytest-xdist lock dir under $TMPDIR in controlled CI.

utilities/copyoffload_migration.py
prabinovRedhat: Agreed — added a DEBUG log with the ValueError message before falling back to the Migration CR lookup, so we keep the original failure context without changing the fallback behavior.

utilities/copyoffload_migration.py
prabinovRedhat: Refactored into _PopulatorConcurrencyTracker with an explicit poll() method and results property. Same behavior, but the polling state is encapsulated and easier to test without nonlocal.

utilities/forklift_controller_populator.py
prabinovRedhat: Good catch — openshift-python-wrapper can deserialize the CR field as a string. Updated the check to int(current_cr_limit) == test_limit so the skip-repatch optimization works when the value is already correct.

utilities/copyoffload_plan_secret.py
```diff
@@ -0,0 +1,45 @@
...
+            wait_timeout=60,
+            sleep=2,
+            func=lambda: any(
+                s.name.startswith(f"{plan_name}-") for s in Secret.get(client=ocp_admin_client, namespace=namespace)
+            ),
+        ):
+            if sample:
+                return
+    except TimeoutExpiredError:
+        LOGGER.warning(f"Timeout waiting for plan secret '{plan_name}-*' - continuing anyway")
```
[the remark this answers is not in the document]
prabinovRedhat: fixed

tests/copyoffload/test_copyoffload_migration.py
```diff
@@ -3472,6 +3478,203 @@ def test_check_vms(
...
+)
+class TestCopyoffloadPopulatorThrottlingMigration:
+    """Copy-offload migration (MTV-696): per-ESXi-host populator throttling and sourceHost labels.
+
+    Covers MTV-5654 Test 1 + Test 3:
+    - Set controller_max_populator_inflight to a low value (2) on ForkliftController
+    - Migrate a VM with 4+ disks from a single ESXi host
+    - Verify peak populator concurrency per host respects the limit and reaches min(limit, disk_count)
+    - Verify PopulatorThrottled events on at least (disk_count − limit) PVCs and sourceHost labels on populate pods
+    - Restore controller_max_populator_inflight to its pre-test value after the class completes
```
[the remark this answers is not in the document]
prabinovRedhat: Replaced the Unicode minus with ASCII - in the docstring to satisfy RUF002 and keep grep/copy-paste predictable.

tests/copyoffload/conftest.py
```diff
@@ -148,6 +162,72 @@ def multi_datastore_config(source_provider_data: dict[str, Any]) -> None:
...
+    This fixture mutates cluster-wide MTV populator settings. Do not run multiple populator
+    throttling test classes against the same cluster in parallel.
+
+    Args:
+        ocp_admin_client (DynamicClient): OpenShift admin client.
+        mtv_namespace (str): Namespace where ForkliftController is installed.
+
+    Yields:
+        None
+    """
```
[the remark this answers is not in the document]
prabinovRedhat: Added the Raises section documenting ValueError when MAX_POPULATOR_INFLIGHT is missing on the deployment and TimeoutError when the cross-worker file lock times out after 3600 seconds.

README.md
```diff
@@ -276,6 +276,12 @@ running the full suite. The sanity set covers:
 | MTV-573 | `TestCopyoffload10MixedDisksMigration` | VM with 10 mixed disks (thin and thick provisioned) |
 | MTV-577 | `TestCopyoffloadWarmMigration` | Warm (live) migration with copy-offload |
 
+> **Note:** `TestCopyoffloadPopulatorThrottlingMigration` (MTV-696) mutates cluster-wide
+> `ForkliftController` populator settings. Do not run it in parallel with other copy-offload tests on the
+> same cluster — the test temporarily sets `controller_max_populator_inflight` (and the populator
+> deployment `MAX_POPULATOR_INFLIGHT` env) to **2**, which caps in-flight populate pods cluster-wide for
+> every other migration running at the same time.
+
```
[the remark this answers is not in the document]
prabinovRedhat: Moved the MTV-696 note from the copyoffload_sanity section to Copy-Offload: Accelerated Migrations (Advanced), since TestCopyoffloadPopulatorThrottlingMigration is not in the sanity marker suite and the warning applies when running the full copyoffload tests on a shared cluster.

utilities/copyoffload_plan_secret.py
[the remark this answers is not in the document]
prabinovRedhat: Updated to re-raise TimeoutExpiredError with plan name and namespace context when the secret gate isn’t met within 60s.

rh-bot-1:
**Clean rebase detected** — no code changes compared to previous head (`79e6dc8`).

tests/copyoffload/test_copyoffload_migration.py
```diff
@@ -3472,6 +3478,203 @@ def test_check_vms(
...
+)
+@pytest.mark.usefixtures(
+    "vmware_cloud_init_ready",
+    "multus_network_name",
+    "copyoffload_config",
+    "populator_inflight_forkliftcontroller",
+    "copyoffload_ssh_key",
+    "cleanup_migrated_vms",
+)
+class TestCopyoffloadPopulatorThrottlingMigration:
```
myakove, line 3499: **Missing AGENTS.md documentation for new 7-step copy-offload throttling pattern**

This PR introduces a new test step pattern:
storagemap → networkmap → plan → migrate → `verify_populator_throttling` → `check_xcopy_used` → `check_vms`

AGENTS.md only documents the 5-step base pattern, 6-step shared-disk pattern, and 6-step copy-offload pattern. The new `test_verify_populator_throttling` step and this 7-step pattern need to be documented in:
- The test structure patterns section
- The test method naming section
- The "Adding New Tests" section

utilities/copyoffload_migration.py
```diff
@@ -10,24 +10,37 @@
...
+from ocp_resources.migration import Migration
 from ocp_resources.pod import Pod
-from ocp_resources.secret import Secret
+from ocp_resources.plan import Plan
+from pytest_testconfig import config as py_config
 from rrmngmnt import Host, RootUser, User
 from simple_logger.logger import get_logger
 from timeout_sampler import TimeoutExpiredError, TimeoutSampler
 
+from exceptions.exceptions import MigrationNotFoundError, MigrationPlanExecError
```
myakove: **Unused import: `MigrationPlanExecError`**

Imported but never used at runtime — only referenced in a docstring (line 858). Docstring references don't require an actual import.

```suggestion
from exceptions.exceptions import MigrationNotFoundError
```

utilities/forklift_controller_populator.py
```diff
@@ -0,0 +1,207 @@
...
+            mtv_namespace=mtv_namespace,
+            expected_limit=test_limit,
+        )
+        yield
+        return
+
+    LOGGER.info(
+        f"Setting ForkliftController controller_max_populator_inflight from {current_cr_limit!r} to {test_limit}"
+    )
+    with ResourceEditor(patches={forklift_controller: {"spec": {"controller_max_populator_inflight": test_limit}}}):
```
myakove: **Deployment restoration wait skipped on test failure**

If the `yield` raises (test failure/error), `@contextmanager` re-raises at `yield` and the post-yield `wait_for_populator_inflight_deployment` (line 207) is skipped. `ResourceEditor` restores the CR, but the async deployment propagation is never verified — the deployment could be in a transitional state when subsequent tests run.

Also: if a previous run crashed during teardown and left `controller_max_populator_inflight` at the test limit, the early-return path (line 175-184) yields without `ResourceEditor` — the original production value is never restored.

Suggestion: wrap yield in try/finally so the deployment wait runs even on failure:
```python
with ResourceEditor(...):
    ...
    try:
        yield
    finally:
        pass  # ResourceEditor exit restores CR

# Post-restore deployment wait always runs
wait_for_populator_inflight_deployment(...)
```
And consider logging a warning on the early-return path noting the limit may be left over from a previous run.

tests/copyoffload/conftest.py
```diff
@@ -148,6 +162,76 @@ def multi_datastore_config(source_provider_data: dict[str, Any]) -> None:
...
+    Raises:
+        ValueError: If MAX_POPULATOR_INFLIGHT is not set on the populator deployment.
+        TimeoutError: If the cross-worker file lock cannot be acquired within 3600 seconds.
+    """
+    lock_path = get_forkliftcontroller_populator_inflight_lock_path()
+    try:
+        with filelock.FileLock(lock_path, timeout=3600):
+            forklift_controller = ForkliftController(
+                client=ocp_admin_client,
+                name="forklift-controller",
```
myakove: **Magic string `"forklift-controller"` not extracted to a constant**

This string also appears at line 624. Per project guidelines on magic values, repeated string literals should be named constants.

Suggestion: Define `FORKLIFT_CONTROLLER_NAME = "forklift-controller"` in `utilities/copyoffload_constants.py` (or similar) and reference it from both locations.

utilities/forklift_controller_populator.py
```diff
@@ -0,0 +1,207 @@
...
+    Args:
+        forklift_controller (ForkliftController): ForkliftController resource to patch.
+        ocp_admin_client (DynamicClient): OpenShift admin client.
+        mtv_namespace (str): Namespace where the populator controller runs.
+        test_limit (int): Limit to apply for the test (e.g. POPULATOR_INFLIGHT_LIMIT).
+        original_deployment_limit (int): MAX_POPULATOR_INFLIGHT value before the test.
+    """
+    current_cr_limit = getattr(forklift_controller.instance.spec, "controller_max_populator_inflight", None)
+
+    if current_cr_limit is not None and int(current_cr_limit) == test_limit:
```
myakove: **`int()` conversion without error handling on Kubernetes API value**

`current_cr_limit` comes from the Kubernetes API via `getattr(forklift_controller.instance.spec, ...)`. If the API returns a non-numeric string (e.g., corrupted CR), `int(current_cr_limit)` raises a raw `ValueError` with no context.

Suggestion:
```python
try:
    cr_limit_int = int(current_cr_limit)
except (ValueError, TypeError) as err:
    raise ValueError(
        f"controller_max_populator_inflight on ForkliftController has non-integer value "
        f"{current_cr_limit!r}"
    ) from err
```

utilities/copyoffload_plan_secret.py
```diff
@@ -0,0 +1,91 @@
...
+    """
+    for secret in Secret.get(client=ocp_admin_client, namespace=namespace):
+        if secret.name.startswith(f"{plan_name}-"):
+            return True
+        labels: dict[str, str] = secret.instance.metadata.labels or {}
+        if labels.get(PLAN_NAME_LABEL) == plan_name and labels.get(POPULATOR_LABEL):
+            return True
+    return False
+
+
```
myakove, line 59: **Plan secret name-prefix matching can produce false positives**

`secret.name.startswith(f"{plan_name}-")` matches secrets belonging to OTHER plans whose names share the same prefix. For example, plan `"auto-abc123-plan"` would match `"auto-abc123-plan-extra-secret"`. The first `return True` short-circuits before the label check on line 46, so a false positive means the wait returns immediately without the correct plan secret existing.

Suggestion: combine both conditions — check name prefix AND labels together, or check labels first (the more precise match):
```python
for secret in Secret.get(client=ocp_admin_client, namespace=namespace):
    labels: dict[str, str] = secret.instance.metadata.labels or {}
    if labels.get(PLAN_NAME_LABEL) == plan_name and labels.get(POPULATOR_LABEL):
        return True
    if secret.name.startswith(f"{plan_name}-"):
        return True
```

myakove (commented):
## Code Review

### :bulb: Suggestions (6)

| File | Line | Issue |
|------|------|-------|
| `tests/copyoffload/test_copyoffload_migration.py` | 3498 | **Missing AGENTS.md documentation for new 7-step copy-offload throttling pattern |
| `utilities/copyoffload_migration.py` | 26 | **Unused import: `MigrationPlanExecError`** |
| `utilities/forklift_controller_populator.py` | 190 | **Deployment restoration wait skipped on test failure** |
| `tests/copyoffload/conftest.py` | 196 | **Magic string `"forklift-controller"` not extracted to a constant** |
| `utilities/forklift_controller_populator.py` | 177 | **`int()` conversion without error handling on Kubernetes API value** |
| `utilities/copyoffload_plan_secret.py` | 44 | **Plan secret name-prefix matching can produce false positives** |

---
*Review generated by pi*

tests/copyoffload/test_copyoffload_migration.py
prabinovRedhat: Addressed in `4aef189`. `AGENTS.md` is a symlink to `CLAUDE.md`; added the 7-step copy-offload throttling pattern to Key Patterns, test method naming (`test_verify_populator_throttling`), and Adding New Tests.

utilities/copyoffload_migration.py
prabinovRedhat: Addressed in `3e0e2e6`. Removed the unused import; the docstring still documents the exception raised by `wait_for_migration_complate()`.

utilities/forklift_controller_populator.py
prabinovRedhat: Addressed in `3e0e2e6`. Wrapped the test body in `try`/`finally` so deployment restoration always runs after failures. Early-return is only used when CR and deployment already match the test limit; otherwise we warn and `_ensure_forklift_controller_populator_limit()` patches the CR back before waiting on the deployment.

tests/copyoffload/conftest.py
prabinovRedhat: Addressed in `3e0e2e6`. Added `FORKLIFT_CONTROLLER_NAME` to `utilities/copyoffload_constants.py` and use it in `tests/copyoffload/conftest.py` and root `conftest.py`.

utilities/forklift_controller_populator.py
prabinovRedhat: Addressed in `3e0e2e6`. Added `_controller_max_populator_inflight_as_int()` with an explicit `ValueError` when the API returns a non-integer value.

utilities/copyoffload_plan_secret.py
prabinovRedhat: Addressed in `3e0e2e6`. `_plan_secret_exists()` now checks `plan-name` + `isPopulator` labels before the `{plan_name}-*` name-prefix fallback.

redhat-qe-bot1:
**Clean rebase detected** — no code changes compared to previous head (`4aef189`).
The following labels were preserved: `commented-prabinovRedhat`, `changes-requested-coderabbitai[bot]`.

utilities/mtv_migration.py
```diff
@@ -347,7 +363,18 @@ def get_plan_migration_status(plan: Plan) -> str:
...
+    *,
+    on_status_poll: Callable[[str], None] | None = None,
+) -> None:
+    """Wait for a Plan migration to reach Succeeded or Failed.
+
+    Args:
+        plan (Plan): The Plan resource to monitor.
+        on_status_poll (Callable[[str], None] | None): Optional callback invoked on each poll
+            with the current migration status string.
+    """
```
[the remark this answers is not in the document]
prabinovRedhat: Addressed in the latest commit. Added a `Raises` section documenting `MigrationPlanExecError` for failed migrations and timeouts, including the `TimeoutSampler` / `py_config['plan_wait_timeout']` polling source.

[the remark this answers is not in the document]
prabinovRedhat: Fixed in `fc4b82d`. Added the `Raises` section to `wait_for_migration_complate()` documenting `MigrationPlanExecError` on migration failure/timeout, including the `TimeoutSampler` polling via `py_config['plan_wait_timeout']`.

prabinovRedhat:
@coderabbitai review

utilities/copyoffload_migration.py
```diff
@@ -691,3 +738,329 @@ def verify_xcopy_used_per_datastore(
...
+            for event in Event.list(
+                client=ocp_admin_client,
+                namespace=target_namespace,
+                field_selector=field_selector,
+                since_seconds=py_config["plan_wait_timeout"],
+            ):
+                if event.get("reason") == POPULATOR_THROTTLED_EVENT_REASON:
+                    seen.add(pvc_name)
+                    LOGGER.info(f"PVC '{pvc_name}' has {POPULATOR_THROTTLED_EVENT_REASON} event")
+                    break
```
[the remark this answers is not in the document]
prabinovRedhat: Fixed in b302963: changed `event.get("reason")` to `event.instance.reason` to match the `.instance` access pattern used consistently throughout the file (e.g. `pod.instance.metadata.labels`, `plan.instance.status`). Also reverted logging back to f-string per project conventions.

utilities/copyoffload_migration.py
```diff
@@ -691,3 +738,329 @@ def verify_xcopy_used_per_datastore(
...
+            field_selector = f"involvedObject.name={pvc_name},involvedObject.kind=PersistentVolumeClaim"
+            for event in Event.list(
+                client=ocp_admin_client,
+                namespace=target_namespace,
+                field_selector=field_selector,
+                since_seconds=py_config["plan_wait_timeout"],
+            ):
+                if event.instance.reason == POPULATOR_THROTTLED_EVENT_REASON:
+                    seen.add(pvc_name)
+                    LOGGER.info("PVC '%s' has %s event", pvc_name, POPULATOR_THROTTLED_EVENT_REASON)
```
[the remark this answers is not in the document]
prabinovRedhat: Fixed in 388334d: reverted to f-string `LOGGER.info(f"PVC '{pvc_name}' has {POPULATOR_THROTTLED_EVENT_REASON} event")` per project logging conventions.

prabinovRedhat:
@myakove — all review threads from June 11–16 are addressed on **`bba1fde`**. Summary:

**Populator fixture / restore**
- `populator_inflight_forkliftcontroller` delegates to `populator_inflight_limit()` in `utilities/forklift_controller_populator.py`
- `try`/`finally` always restores `controller_max_populator_inflight` to the pre-test deployment value and waits for the populator deployment to reconcile, including on test failure
- CR limit comparisons use `_controller_max_populator_inflight_as_int()` (string `"2"` vs int `2` handled)
- Helpers moved out of conftest; `FORKLIFT_CONTROLLER_NAME`, lock-dir hardening (`O_NOFOLLOW`), etc. — see replies on each thread

**Other items**
- 7-step throttling pattern documented in `CLAUDE.md` / `AGENTS.md`
- Throttling test uses `prepared_plan["copyoffload"]` (fail-fast)
- Jenkins tox is green; CodeRabbit approved `bba1fde`

**Plan secret wait (best-effort WARNING)**
The `wait_for_plan_secret()` timeout warning is expected: Forklift creates populator/vm-config secrets at **migration** time, not Plan Ready, so the 60s wait in `test_create_plan` routinely times out without indicating a real failure.

This is **intentionally out of scope for this PR**. Follow-up is tracked in **[MTV-5799](https://redhat.atlassian.net/browse/MTV-5799)** — remove or relocate `wait_for_plan_secret()` in a separate PR so we stop adding 60s + log noise on every copy-offload run.

utilities/forklift_controller_populator.py
```diff
@@ -0,0 +1,273 @@
...
+            test_limit,
+            original_deployment_limit,
+        )
+
+    LOGGER.info(
+        f"Setting ForkliftController controller_max_populator_inflight from {current_cr_limit!r} to {test_limit}"
+    )
+    try:
+        with ResourceEditor(patches={forklift_controller: {"spec": {"controller_max_populator_inflight": test_limit}}}):
+            forklift_controller.wait_for_condition(
```
myakove: **Double-restore: ResourceEditor auto-restores, then `finally` block restores again**

`ResourceEditor` as a context manager auto-restores the CR on `__exit__`, then the `finally` block calls `_ensure_forklift_controller_populator_limit()` which patches the CR again. This creates two successive patches (two reconcile cycles).

If the CR's pre-test value differs from the deployment env var (e.g., operator drift, previous crashed run), the `finally` block overwrites ResourceEditor's restoration with the deployment value — which may be the correct intent, but it's not documented.

Suggestion: use `ResourceEditor(...).update(backup_resources=False)` + explicit restore in `finally` to avoid the double-patch, since you already handle restore explicitly. Or document the intentional double-restore behavior.

utilities/forklift_controller_populator.py
```diff
@@ -0,0 +1,273 @@
...
+        available_replicas = 0
+        if deployment_status:
+            available_replicas = deployment_status.availableReplicas or 0
+        if available_replicas < spec_replicas:
+            return False
+        return get_populator_inflight_from_deployment(deployment=current_deployment) == expected_value
+
+    try:
+        for ready in TimeoutSampler(
+            wait_timeout=300,
```
myakove: **Magic number `300` used 3 times without a named constant**

The timeout value `300` appears at lines 154, 202, and 256 — all for ForkliftController condition waits and deployment readiness checks. Per project guidelines on magic numbers, extract a named constant:

```python
_FORKLIFT_CONTROLLER_CONDITION_TIMEOUT = 300  # seconds to wait for ForkliftController reconciliation
```

tests/copyoffload/conftest.py
```diff
@@ -148,6 +164,76 @@ def multi_datastore_config(source_provider_data: dict[str, Any]) -> None:
     LOGGER.info("✓ Multi-datastore configuration validated: secondary_datastore_id = %s", secondary_datastore_id)
 
 
+@pytest.fixture(scope="class")
+def populator_inflight_forkliftcontroller(
```
myakove, line 166: **Fixture `populator_inflight_forkliftcontroller` — consider extracting deployment-limit-reading helper**

The fixture body is ~40 lines of logic (plus ~18 lines docstring) handling: lock acquisition, ForkliftController construction/validation, deployment limit reading, and delegation to the context manager. The deployment-limit-reading block (lines 210-225) is a separable concern.

Suggestion: extract to a small helper like `_get_current_deployment_populator_limit(ocp_admin_client, mtv_namespace)` for readability.

utilities/forklift_controller_populator.py
```diff
@@ -0,0 +1,273 @@
...
+    )
+    forklift_controller.wait_for_condition(
+        status=forklift_controller.Condition.Status.TRUE,
+        condition=forklift_controller.Condition.Type.SUCCESSFUL,
+        timeout=300,
+    )
+
+
+@contextmanager
+def populator_inflight_limit(
```
myakove, line 280: **`populator_inflight_limit` at 67 lines — consider splitting**

This function has three distinct paths: already-at-limit early return, CR-mismatch warning + patch, and the main patch-with-restore flow. Per the 50-line guideline, consider extracting the early-return validation block (lines 229-237) or adding section comments to delineate the paths.

utilities/copyoffload_migration.py
```diff
@@ -691,3 +737,354 @@ def verify_xcopy_used_per_datastore(
...
+    """
+    event_resource = ocp_admin_client.resources.get(api_version=Event.api_version, kind="Event")
+    response = event_resource.get(
+        namespace=namespace,
+        field_selector=f"involvedObject.name={pvc_name},involvedObject.kind=PersistentVolumeClaim",
+    )
+    return response.items or []
+
+
+def _verify_throttled_events_on_pods(
```
myakove: **`_verify_throttled_events_on_pods` at 68 lines — consider extracting PVC name collection**

The PVC name collection logic (lines ~963-977) and the event-scanning loop (lines ~979-988) are separable units that could be extracted into small helpers, bringing the function under the 50-line guideline.

utilities/forklift_controller_populator.py
```diff
@@ -0,0 +1,273 @@
...
+    test failures.
+
+    Args:
+        forklift_controller (ForkliftController): ForkliftController resource to patch.
+        ocp_admin_client (DynamicClient): OpenShift admin client.
+        mtv_namespace (str): Namespace where the populator controller runs.
+        test_limit (int): Limit to apply for the test (e.g. POPULATOR_INFLIGHT_LIMIT).
+        original_deployment_limit (int): MAX_POPULATOR_INFLIGHT value before the test.
+    """
+    current_cr_limit = getattr(forklift_controller.instance.spec, "controller_max_populator_inflight", None)
```
myakove: **Duplicated `getattr` expression for `controller_max_populator_inflight`**

`getattr(forklift_controller.instance.spec, "controller_max_populator_inflight", None)` appears at lines 192 and 227 (in two different functions). Per "No Duplicate Code" (MUST): same expression computed multiple times should be stored/extracted.

Suggestion: extract a helper like:
```python
def _get_cr_populator_limit(forklift_controller: ForkliftController) -> int | None:
    raw = getattr(forklift_controller.instance.spec, "controller_max_populator_inflight", None)
    return _controller_max_populator_inflight_as_int(raw)
```
Both call sites already follow the `getattr` → `_controller_max_populator_inflight_as_int()` pattern.

myakove (commented):
## Code Review

### :bulb: Suggestions (6)

| File | Line | Issue |
|------|------|-------|
| `utilities/forklift_controller_populator.py` | 253 | **Double-restore: ResourceEditor auto-restores, then `finally` block restores ag |
| `utilities/forklift_controller_populator.py` | 154 | **Magic number `300` used 3 times without a named constant** |
| `tests/copyoffload/conftest.py` | 168 | **Fixture `populator_inflight_forkliftcontroller` — consider extracting deployme |
| `utilities/forklift_controller_populator.py` | 207 | **`populator_inflight_limit` at 67 lines — consider splitting** |
| `utilities/copyoffload_migration.py` | 935 | **`_verify_throttled_events_on_pods` at 68 lines — consider extracting PVC name  |
| `utilities/forklift_controller_populator.py` | 227 | **Duplicated `getattr` expression for `controller_max_populator_inflight`** |

---
*Review generated by pi*

utilities/forklift_controller_populator.py
prabinovRedhat: Fixed in 94aa796. Setup now uses `_ensure_forklift_controller_populator_limit(..., backup_resources=False)` without a `ResourceEditor` context manager. Restore happens only once in the `finally` block. Verified locally: single CR patch 20→2 on setup and 2→20 on teardown (deployment revision 55→56→57).

utilities/forklift_controller_populator.py
prabinovRedhat: Fixed in 94aa796. Extracted `FORKLIFT_CONTROLLER_CONDITION_TIMEOUT = 300` and replaced all three hardcoded `300` usages (condition wait, deployment readiness sampler, and fixture RUNNING wait).

tests/copyoffload/conftest.py
prabinovRedhat: Fixed in 94aa796. Added `get_deployment_populator_inflight_limit()` in `utilities/forklift_controller_populator.py`; the fixture now delegates deployment limit reading and validation to that helper.

utilities/forklift_controller_populator.py
prabinovRedhat: Addressed in 94aa796. Split validation into `_get_cr_populator_limit()`, `_warn_if_cr_limit_leftover_from_crashed_run()`, and kept restore in a single `finally` block. `populator_inflight_limit` is now shorter and each helper has a single responsibility.

utilities/copyoffload_migration.py
prabinovRedhat: Fixed in 94aa796. Extracted `_collect_migration_pvc_names()` and `_find_throttled_pvc_names()`; `_verify_throttled_events_on_pods` now orchestrates validation only.

utilities/forklift_controller_populator.py
prabinovRedhat: Fixed in 94aa796. Added `_get_cr_populator_limit()` and use it in both `_ensure_forklift_controller_populator_limit` and `populator_inflight_limit` instead of duplicating `getattr(...)`.

prabinovRedhat:
@myakove Replied in each of your June 18 review threads with details for commit `94aa796`. Summary:

- **Double-restore** → single `finally` restore with `backup_resources=False`; verified one patch each way locally
- **Magic 300** → `FORKLIFT_CONTROLLER_CONDITION_TIMEOUT`
- **Fixture helper** → `get_deployment_populator_inflight_limit()`
- **`populator_inflight_limit` length** → split into `_get_cr_populator_limit`, `_warn_if_cr_limit_leftover_from_crashed_run`
- **`_verify_throttled_events_on_pods` length** → `_collect_migration_pvc_names`, `_find_throttled_pvc_names`
- **Duplicated getattr** → `_get_cr_populator_limit()`

Local run after these changes: **7/7 passed** (~17m). Cluster restore confirmed (CR + deployment back to 20).

utilities/copyoffload_plan_secret.py
```diff
@@ -0,0 +1,91 @@
...
+            func=lambda: _plan_secret_exists(
+                ocp_admin_client=ocp_admin_client,
+                namespace=namespace,
+                plan_name=plan_name,
+            ),
+        ):
+            if sample:
+                return
+    except TimeoutExpiredError:
+        secret_names = _list_namespace_secret_names(ocp_admin_client=ocp_admin_client, namespace=namespace)
```
myakove, line 101: **Silent recovery on timeout violates "No Fallbacks, No Silent Recovery" (MUST)**

AGENTS.md: "When something is wrong, fail immediately — never fall back to an alternative." This catches `TimeoutExpiredError` and continues with `LOGGER.warning`. The docstring says "best-effort" but the project rules explicitly prohibit this pattern.

Suggestion: Raise `TimeoutError` instead of continuing. If the intent is to let migration produce a clearer error, add an explicit code comment referencing the justification for the exception to the rule.

utilities/forklift_controller_populator.py
```diff
@@ -0,0 +1,336 @@
...
+    finally:
+        _ensure_forklift_controller_populator_limit(
+            forklift_controller=forklift_controller,
+            target_limit=original_deployment_limit,
+        )
+        wait_for_populator_inflight_deployment(
+            ocp_admin_client=ocp_admin_client,
+            mtv_namespace=mtv_namespace,
+            expected_limit=original_deployment_limit,
+        )
```
myakove: **`finally` block can mask original test exception**

If a test fails (exception at `yield`) and then `_ensure_forklift_controller_populator_limit` or `wait_for_populator_inflight_deployment` in this `finally` block also raises (e.g., `TimeoutError`), the original test failure is replaced by the cleanup error. The caller never sees what actually went wrong.

Suggestion:
```python
    finally:
        try:
            _ensure_forklift_controller_populator_limit(
                forklift_controller=forklift_controller,
                target_limit=original_deployment_limit,
            )
            wait_for_populator_inflight_deployment(
                ocp_admin_client=ocp_admin_client,
                mtv_namespace=mtv_namespace,
                expected_limit=original_deployment_limit,
            )
        except Exception:
            LOGGER.exception(
                "Failed to restore ForkliftController populator limit to %s during cleanup",
                original_deployment_limit,
            )
```

myakove (commented):
## Code Review

### :bulb: Suggestions (5)

| File | Line | Issue |
|------|------|-------|
| `utilities/copyoffload_plan_secret.py` | 87 | **Silent recovery on timeout violates "No Fallbacks, No Silent Recovery" (MUST)* |
| `utilities/forklift_controller_populator.py` | 336 | **`finally` block can mask original test exception** |
| `utilities/forklift_controller_populator.py` | 206 | **Bare `int()` without descriptive error context — inconsistent with `_controlle |
| `tests/copyoffload/test_copyoffload_migration.py` | 3639 | **Redundant logging in test method — `verify_populator_throttling()` already log |
| `utilities/forklift_controller_populator.py` | 239 | **Parameterized `%s` logging for non-expensive values** |

---
*Review generated by pi*

utilities/forklift_controller_populator.py
```diff
@@ -0,0 +1,336 @@
...
+        name=POPULATOR_CONTROLLER_DEPLOYMENT,
+        namespace=mtv_namespace,
+        ensure_exists=True,
+    )
+    limit_str = get_populator_inflight_from_deployment(deployment=deployment)
+    if limit_str is None:
+        raise ValueError(
+            f"{MAX_POPULATOR_INFLIGHT_ENV} not found on {POPULATOR_CONTROLLER_DEPLOYMENT} "
+            f"before populator throttling test setup"
+        )
```
myakove, line 207: **Bare `int()` without descriptive error context — inconsistent with `_controller_max_populator_inflight_as_int`**

Compare with `_controller_max_populator_inflight_as_int` (line 37-48) which wraps `int()` conversion with a clear contextual error. This bare `int(limit_str)` gives an unhelpful `ValueError: invalid literal for int() with base 10: 'abc'`.

Suggestion: reuse the existing helper or wrap with context:
```python
try:
    return int(limit_str)
except (ValueError, TypeError) as err:
    raise ValueError(
        f"{MAX_POPULATOR_INFLIGHT_ENV} on {POPULATOR_CONTROLLER_DEPLOYMENT} has non-integer value {limit_str!r}"
    ) from err
```

tests/copyoffload/test_copyoffload_migration.py
```diff
@@ -3472,6 +3478,203 @@ def test_check_vms(
...
+    ) -> None:
+        """Verify sourceHost labels, PopulatorThrottled events, and observed concurrency."""
+        source_host = verify_populator_throttling(
+            ocp_admin_client=ocp_admin_client,
+            plan=self.plan_resource,
+            target_namespace=target_namespace,
+            max_concurrent_by_host=self.max_concurrent_by_host,
+            max_populator_inflight=POPULATOR_INFLIGHT_LIMIT,
+        )
+        LOGGER.info(f"All populate pods share sourceHost={source_host!r}")
```
myakove: **Redundant logging in test method — `verify_populator_throttling()` already logs all details**

Per "No Unnecessary Logging" (MUST): "Reporter/progress logging in test methods" should be flagged. The called function logs per-pod sourceHost labels, throttling results, and concurrency observations internally.

Suggestion: Remove this `LOGGER.info(...)` line.

utilities/forklift_controller_populator.py
```diff
@@ -0,0 +1,336 @@
...
+) -> None:
+    """Log when the CR limit matches the test value but the deployment reports a different limit.
+
+    Args:
+        cr_limit_int (int | None): Parsed CR limit before patching.
+        test_limit (int): Limit applied for the test.
+        original_deployment_limit (int): MAX_POPULATOR_INFLIGHT value before the test.
+    """
+    if cr_limit_int == test_limit:
+        LOGGER.warning(
```
myakove, line 245: **Parameterized `%s` logging for non-expensive values**

Per AGENTS.md Logging Format: "Use f-strings for logging by default. Use parameterized format (`%s`) only for expensive operations." Simple integers are not expensive.

Suggestion:
```python
        LOGGER.warning(
            f"ForkliftController controller_max_populator_inflight already at test limit {test_limit} "
            f"but deployment reports {original_deployment_limit}; "
            "this may be leftover from a previous crashed run"
        )
```

utilities/copyoffload_plan_secret.py
prabinovRedhat: Addressed in 6d3797f. Raising `TimeoutError` here would fail every copy-offload test at Plan Ready because Forklift creates populator secrets at **migration start**, not Plan Ready (see MTV-5799). Documented this intentional exception in the docstring and inline comment; MTV-5799 tracks moving `wait_for_plan_secret` to migration start, after which timeout will fail fast.

prabinovRedhat:
@myakove Addressed your latest 5 review threads in `6d3797f` — replies posted in each thread. Summary:

1. **Plan secret timeout** — documented MTV-5799 exception (secret appears at migration start, not Plan Ready)
2. **Finally masking test failure** — cleanup errors logged; only re-raised when no active test exception
3. **Bare int()** — contextual ValueError in `get_deployment_populator_inflight_limit()`
4. **Redundant test logging** — removed duplicate LOGGER.info
5. **%s logging** — f-string in `_warn_if_cr_limit_leftover_from_crashed_run`

utilities/forklift_controller_populator.py
prabinovRedhat: Fixed in 6d3797f. Wrapped restore cleanup in `try/except (TimeoutError, ValueError)` with `LOGGER.exception`. Cleanup errors are re-raised only when no test exception is active (`sys.exc_info()[0] is None`), so a failing test is not masked by a restore timeout.

utilities/forklift_controller_populator.py
prabinovRedhat: Fixed in 6d3797f. `get_deployment_populator_inflight_limit()` now wraps `int(limit_str)` with the same contextual `ValueError` pattern used elsewhere in this module.

tests/copyoffload/test_copyoffload_migration.py
prabinovRedhat: Fixed in 6d3797f. Removed the redundant `LOGGER.info` from `test_verify_populator_throttling`; per-pod `sourceHost` logging remains in `_verify_source_host_labels_on_pods()`.

utilities/forklift_controller_populator.py
prabinovRedhat: Fixed in 6d3797f. Switched `_warn_if_cr_limit_leftover_from_crashed_run` to f-string logging.

utilities/copyoffload_plan_secret.py
prabinovRedhat: fixing in this PR

prabinovRedhat: Fixed in `6cb4755` (MTV-5799):

- Removed plan-secret wait from `create_plan_resource()` (Plan Ready was too early — Forklift creates the populator secret when migration starts).
- Added `wait_for_copyoffload_plan_secret()` after Migration CR creation in `execute_migration()` and `execute_migration_monitoring_populator_inflight()`.
- Removed `"continuing anyway"` silent recovery; timeout now raises `TimeoutError` with secret diagnostics.
- Simultaneous/concurrent tests: create all Migration CRs first, then wait for copy-offload secrets.

Verified locally:
- `TestCopyoffloadPopulatorThrottlingMigration` — secret wait log in `test_migrate_vms`, 7/7 passed
- `TestCopyoffloadThinMigration` — same, 6/6 passed

prabinovRedhat:
**MTV-5799 plan secret wait — fixed in `6cb4755`**

Moved the copy-offload plan populator secret wait from Plan Ready to **after Migration CR creation**, where Forklift actually creates the secret. Removed the `"continuing anyway"` warning path; we now fail fast with `TimeoutError` if the secret does not appear within 60s.

Coverage:
- Standard copy-offload migrations via `execute_migration()`
- Populator throttling via `execute_migration_monitoring_populator_inflight()`
- Simultaneous / XCOPY+VDDK concurrent tests (explicit wait after all Migration CRs exist)

Local runs: thin (6/6) and populator throttling (7/7) — secret wait log appears in `test_migrate_vms`, not `test_create_plan`.

myakove (approved):

redhat-qe-bot2:

PR #552, opened 2026-06-14
feat: PVC-based shared disk device detection

MiriSafra:
## Summary
- Refactor shared disk device detection from source-side heuristic to
  destination-side PVC lookup
- Find the shared disk by intersecting PVC references from both destination
  VMIs — no assumptions about controller layout or disk ordering
- Per-VM device paths handle VMs with different disk counts correctly
- More dynamic: reads what Forklift actually created rather than predicting
  from source config

## Summary by CodeRabbit

* **Tests**
  * Enhanced shared-disk migration verification to use administrative client context for more reliable validation.

* **Improvements**
  * Improved shared-disk device and partition detection by resolving shared PVCs and device targets from each destination virtual machine’s runtime configuration, rather than relying on the source disk layout.
  * Updated verification flow to mount, remount, and flush buffers using per-VM device/partition information for more accurate shared-disk checks.

### DIFF
--- a/utilities/shared_disk.py
+++ b/utilities/shared_disk.py
@@ -137,21 +137,24 @@ def _get_pvc_device_targets(
         raise ValueError(f"VM '{vm_name}' in '{target_namespace}' has no running VMI")
 
     vmi = cnv_vm.vmi
+    pvc_devices: dict[str, str] = {}
     try:
         for sample in TimeoutSampler(wait_timeout=300, sleep=5, func=lambda: vmi.instance):
-            if getattr(sample.status, "volumeStatus", None):
+            volume_status = getattr(sample.status, "volumeStatus", None)
+            if not volume_status:
+                continue
+            pvc_devices.clear()
+            for vol_status in volume_status:
+                pvc_info = getattr(vol_status, "persistentVolumeClaimInfo", None)
+                if pvc_info and vol_status.target:
+                    pvc_devices[pvc_info.claimName] = f"/dev/{vol_status.target}"
+            if pvc_devices:
                 break
     except TimeoutExpiredError:
         raise ValueError(
-            f"VM '{vm_name}' in '{target_namespace}' VMI has no volumeStatus after 300s "
+            f"VM '{vm_name}' in '{target_namespace}' VMI has no PVC device targets after 300s "
             f"(phase: {getattr(vmi.instance.status, 'phase', 'unknown')})"
         ) from None
-
-    pvc_devices: dict[str, str] = {}
-    for vol_status in vmi.instance.status.volumeStatus:
-        pvc_info = getattr(vol_status, "persistentVolumeClaimInfo", None)
-        if pvc_info and vol_status.target:
-            pvc_devices[pvc_info.claimName] = f"/dev/{vol_status.target}"
     return pvc_devices
 
 
--- a/utilities/shared_disk.py
+++ b/utilities/shared_disk.py
@@ -169,7 +169,7 @@ def _get_shared_disk_devices(
     Finds the PVC referenced by both destination VMs (the shared disk) and
     returns each VM's device path based on the PVC's position in that VM's
     full volume list (including non-PVC volumes). Handles VMs with different
-    disk layouts correctly. Works for any bus type (SCSI, IDE, virtio).
+    disk layouts correctly. Assumes virtio bus (Forklift default after migration).
 
     Args:
         ocp_admin_client (DynamicClient): OpenShift admin client.
--- a/utilities/shared_disk.py
+++ b/utilities/shared_disk.py
@@ -7,13 +7,18 @@ after migration.
 from __future__ import annotations
 
 import shlex
-from typing import Any
+from typing import TYPE_CHECKING, Any
 
+from ocp_resources.virtual_machine import VirtualMachine
 from simple_logger.logger import get_logger
 
 from exceptions.exceptions import GuestCommandError
+from utilities.naming import resolve_destination_vm_name
 from utilities.post_migration import get_ssh_credentials_from_provider_config
 from utilities.ssh_utils import SSHConnectionManager, VMSSHConnection
+
+if TYPE_CHECKING:
+    from kubernetes.dynamic import DynamicClient
 
 LOGGER = get_logger(name=__name__)
 
@@ -97,48 +102,107 @@ def _write_marker(ssh_conn: VMSSHConnection, file_path: str, content: str, vm_la
     _run_cmd_on_vm(ssh_conn, ["sudo", "sync"], f"{vm_label} sync")
 
 
-def _get_shared_disk_device(prepared_plan: dict[str, Any], vm_name: str) -> str:
-    """Determine the shared disk device path from source VM disk layout.
-
-    The shared disk is on a separate SCSI controller (Multi-writer requires
-    bus sharing). After migration, KubeVirt assigns /dev/vdX in the same
-    order as the source SCSI layout.
-
-    Note:
-        Assumes exactly two SCSI controllers with the shared disk alone on the
-        non-boot controller. This matches the current test VM layout. If VMs
-        gain additional controllers with non-shared disks, this logic would need
-        to match by exact SCSI position instead.
+def _get_pvc_names_from_vm(
+    ocp_admin_client: "DynamicClient",
+    target_namespace: str,
+    vm_name: str,
+) -> list[str]:
+    """Extract ordered PVC claim names from a destination VM's running instance.
 
     Args:
-        prepared_plan (dict[str, Any]): Plan config with source_vms_data.
-        vm_name (str): Name of the VM to find the shared disk for.
+        ocp_admin_client (DynamicClient): OpenShift admin client.
+        target_namespace (str): Namespace where the migrated VM lives.
+        vm_name (str): Destination VM name (already sanitized for Kubernetes).
 
     Returns:
-        str: Device path (e.g., "/dev/vdc").
+        list[str]: PVC claim names in volume order (cloud-init excluded).
 
     Raises:
-        ValueError: If no shared disk found (no disk on a different controller).
+        ValueError: If the VM has no running instance (VMI).
     """
-    disks: list[dict[str, Any]] = prepared_plan["source_vms_data"][vm_name]["disks"]
-    if not disks:
-        raise ValueError(f"No disks found for VM '{vm_name}' in source_vms_data")
-    # Shared disk is on a separate SCSI controller (Multi-writer requires bus sharing).
-    # Assumes no non-shared disks on separate controllers — valid for current test VM layout.
-    sorted_disks = sorted(disks, key=lambda d: (d["controller_key"], d["unit_number"]))
-    boot_controller: int = sorted_disks[0]["controller_key"]
-    for index, disk in enumerate(sorted_disks):
-        if disk["controller_key"] != boot_controller:
-            device_path = f"/dev/vd{chr(ord('a') + index)}"
-            LOGGER.info(f"Detected shared disk for VM '{vm_name}' at index {index}: {device_path}")
-            return device_path
-    raise ValueError(f"No shared disk found for VM '{vm_name}' — all disks on same SCSI controller")
+    cnv_vm = VirtualMachine(
+        client=ocp_admin_client,
+        name=vm_name,
+        namespace=target_namespace,
+        ensure_exists=True,
+    )
+    if not cnv_vm.vmi:
+        raise ValueError(f"VM '{vm_name}' in '{target_namespace}' has no running VMI")
+
+    pvc_names: list[str] = []
+    for vol in cnv_vm.vmi.instance.spec.volumes:
+        pvc_ref = getattr(vol, "persistentVolumeClaim", None)
+        if pvc_ref:
+            pvc_names.append(pvc_ref.claimName)
+    return pvc_names
+
+
+def _index_to_device_path(idx: int) -> str:
+    """Convert a zero-based volume index to a Linux device path.
+
+    Args:
+        idx (int): Zero-based volume index.
+
+    Returns:
+        str: Device path (e.g., 0 -> "/dev/vda", 2 -> "/dev/vdc").
+    """
+    return f"/dev/vd{chr(ord('a') + idx)}"
+
+
+def _get_shared_disk_devices(
+    ocp_admin_client: "DynamicClient",
+    target_namespace: str,
+    vm1_config: dict[str, Any],
+    vm2_config: dict[str, Any],
+) -> tuple[str, str]:
+    """Determine per-VM shared disk device paths from destination PVC references.
+
+    Finds the PVC referenced by both destination VMs (the shared disk) and
+    returns each VM's device path based on the PVC's position in that VM's
+    volume list. Handles VMs with different disk layouts correctly.
+    Works for any bus type (SCSI, IDE, virtio).
+
+    Args:
+        ocp_admin_client (DynamicClient): OpenShift admin client.
+        target_namespace (str): Namespace where migrated VMs live.
+        vm1_config (dict[str, Any]): First VM config dict from prepared_plan["virtual_machines"].
+        vm2_config (dict[str, Any]): Second VM config dict from prepared_plan["virtual_machines"].
+
+    Returns:
+        tuple[str, str]: Device paths for (VM1, VM2), e.g. ("/dev/vdc", "/dev/vdb").
+
+    Raises:
+        ValueError: If no shared PVC found between the two VMs.
+    """
+    vm1_dest_name = resolve_destination_vm_name(vm1_config)
+    vm2_dest_name = resolve_destination_vm_name(vm2_config)
+
+    vm1_pvcs = _get_pvc_names_from_vm(ocp_admin_client, target_namespace, vm1_dest_name)
+    vm2_pvcs = _get_pvc_names_from_vm(ocp_admin_client, target_namespace, vm2_dest_name)
+
+    shared_pvcs = set(vm1_pvcs) & set(vm2_pvcs)
+    if not shared_pvcs:
+        raise ValueError(
+            f"No shared PVC between '{vm1_dest_name}' and '{vm2_dest_name}'. VM1 PVCs: {vm1_pvcs}, VM2 PVCs: {vm2_pvcs}"
+        )
+
+    shared_pvc = shared_pvcs.pop()
+    vm1_idx = vm1_pvcs.index(shared_pvc)
+    vm2_idx = vm2_pvcs.index(shared_pvc)
+    vm1_device = _index_to_device_path(vm1_idx)
+    vm2_device = _index_to_device_path(vm2_idx)
+    LOGGER.info(
+        f"Shared PVC '{shared_pvc}': '{vm1_dest_name}' index {vm1_idx} -> {vm1_device}, "
+        f"'{vm2_dest_name}' index {vm2_idx} -> {vm2_device}"
+    )
+    return vm1_device, vm2_device
 
 
 def verify_shared_disk_data(
     prepared_plan: dict[str, Any],
     vm_ssh_connections: SSHConnectionManager,
     source_provider_data: dict[str, Any],
+    ocp_admin_client: "DynamicClient",
 ) -> None:
     """Verify shared disk is accessible from both VMs by writing and reading data.
 
@@ -151,17 +215,21 @@ def verify_shared_disk_data(
     3. VM1: flush block device cache, remount, read VM2's data, unmount
 
     Args:
-        prepared_plan (dict[str, Any]): Plan config with virtual_machines and source_vms_data.
+        prepared_plan (dict[str, Any]): Plan config with virtual_machines, source_vms_data, and _vm_target_namespace.
         vm_ssh_connections (SSHConnectionManager): SSH connection manager.
         source_provider_data (dict[str, Any]): Provider configuration from .providers.json.
+        ocp_admin_client (DynamicClient): OpenShift admin client for destination VM lookup.
 
     Raises:
         AssertionError: If shared disk data verification fails.
         GuestCommandError: If SSH commands fail.
     """
-    vm1_name = prepared_plan["virtual_machines"][0]["name"]
-    vm2_name = prepared_plan["virtual_machines"][1]["name"]
-    shared_disk_device = _get_shared_disk_device(prepared_plan, vm1_name)
+    vm1_config = prepared_plan["virtual_machines"][0]
+    vm2_config = prepared_plan["virtual_machines"][1]
+    vm1_name = vm1_config["name"]
+    vm2_name = vm2_config["name"]
+    vm_namespace = prepared_plan["_vm_target_namespace"]
+    vm1_device, vm2_device = _get_shared_disk_devices(ocp_admin_client, vm_namespace, vm1_config, vm2_config)
 
     LOGGER.info(f"Verifying shared disk between {vm1_name} and {vm2_name}")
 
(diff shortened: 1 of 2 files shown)


utilities/shared_disk.py
```diff
@@ -97,48 +102,107 @@ def _write_marker(ssh_conn: VMSSHConnection, file_path: str, content: str, vm_la
...
+    )
+    if not cnv_vm.vmi:
+        raise ValueError(f"VM '{vm_name}' in '{target_namespace}' has no running VMI")
+
+    pvc_names: list[str] = []
+    for vol in cnv_vm.vmi.instance.spec.volumes:
+        pvc_ref = getattr(vol, "persistentVolumeClaim", None)
+        if pvc_ref:
+            pvc_names.append(pvc_ref.claimName)
+    return pvc_names
```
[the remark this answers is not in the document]
MiriSafra: Fixed — removed the inaccurate "(cloud-init excluded)" from the docstring. The filtering is implicit (only `persistentVolumeClaim` volumes are collected), so the parenthetical was misleading.

utilities/shared_disk.py
```diff
@@ -97,48 +102,107 @@ def _write_marker(ssh_conn: VMSSHConnection, file_path: str, content: str, vm_la
...
+    shared_pvc = shared_pvcs.pop()
+    vm1_idx = vm1_pvcs.index(shared_pvc)
+    vm2_idx = vm2_pvcs.index(shared_pvc)
+    vm1_device = _index_to_device_path(vm1_idx)
+    vm2_device = _index_to_device_path(vm2_idx)
+    LOGGER.info(
+        f"Shared PVC '{shared_pvc}': '{vm1_dest_name}' index {vm1_idx} -> {vm1_device}, "
+        f"'{vm2_dest_name}' index {vm2_idx} -> {vm2_device}"
+    )
+    return vm1_device, vm2_device
```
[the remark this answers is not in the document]
MiriSafra: Good catch — added the fail-fast guard for `len(shared_pvcs) > 1`. Kept `.pop()` instead of `next(iter(...))` since after the guard we're guaranteed exactly one element, so both are equivalent.

utilities/shared_disk.py
```diff
@@ -97,48 +102,112 @@ def _write_marker(ssh_conn: VMSSHConnection, file_path: str, content: str, vm_la
...
+    volume list. Handles VMs with different disk layouts correctly.
+    Works for any bus type (SCSI, IDE, virtio).
+
+    Args:
+        ocp_admin_client (DynamicClient): OpenShift admin client.
+        target_namespace (str): Namespace where migrated VMs live.
+        vm1_config (dict[str, Any]): First VM config dict from prepared_plan["virtual_machines"].
+        vm2_config (dict[str, Any]): Second VM config dict from prepared_plan["virtual_machines"].
+
+    Returns:
```
myakove, line 194: better to return here dict which is more readable

myakove (changes requested):

utilities/shared_disk.py
```diff
@@ -97,48 +102,112 @@ def _write_marker(ssh_conn: VMSSHConnection, file_path: str, content: str, vm_la
...
+    """
+    cnv_vm = VirtualMachine(
+        client=ocp_admin_client,
+        name=vm_name,
+        namespace=target_namespace,
+        ensure_exists=True,
+    )
+    if not cnv_vm.vmi:
+        raise ValueError(f"VM '{vm_name}' in '{target_namespace}' has no running VMI")
+
```
myakove: **Potential device path mismatch:** `_get_pvc_names_from_vm` collects only PVC-backed volumes from `spec.volumes`, then `_index_to_device_path` maps that PVC-only index to `/dev/vdX`.

But KubeVirt assigns device letters based on **all** volumes/disks — including non-PVC ones like `cloudinit`. If a VM has volumes `[cloudinit, pvc-boot, pvc-shared]`, the PVC-only list is `[pvc-boot, pvc-shared]` with indices 0,1 → `/dev/vda`, `/dev/vdb`. But the actual devices would be `/dev/vda`=cloudinit, `/dev/vdb`=pvc-boot, `/dev/vdc`=pvc-shared — causing the test to mount the wrong device.

Consider iterating ALL volumes (tracking each volume's position regardless of type), or using `spec.domain.devices.disks` which has explicit device name assignments.

utilities/shared_disk.py
myakove, line 194: Previous review comment still unresolved: consider returning a `dict` instead of `tuple[str, str]` for better readability at the call site — e.g. `{"vm1": "/dev/vdc", "vm2": "/dev/vdb"}` makes it clear which device belongs to which VM, vs positional tuple unpacking.

utilities/shared_disk.py
```diff
@@ -97,48 +102,112 @@ def _write_marker(ssh_conn: VMSSHConnection, file_path: str, content: str, vm_la
...
 
     Args:
-        prepared_plan (dict[str, Any]): Plan config with source_vms_data.
-        vm_name (str): Name of the VM to find the shared disk for.
+        idx (int): Zero-based volume index.
 
     Returns:
-        str: Device path (e.g., "/dev/vdc").
+        str: Device path (e.g., 0 -> "/dev/vda", 2 -> "/dev/vdc").
+    """
```
myakove: No bounds check — if `idx >= 26`, `chr(ord('a') + idx)` produces non-letter characters (e.g., `{`, `|`). Practically unlikely with current VM configs, but a defensive check would give a clearer error:

```python
def _index_to_device_path(idx: int) -> str:
    if idx > 25:
        raise ValueError(f"Volume index {idx} exceeds supported device range (a-z)")
    return f"/dev/vd{chr(ord('a') + idx)}"
```

myakove (commented):
## Code Review

### :bulb: Suggestions (3)

| File | Line | Issue |
|------|------|-------|
| `utilities/shared_disk.py` | 131 | **Potential device path mismatch:** `_get_pvc_names_from_vm` collects only PVC-b |
| `utilities/shared_disk.py` | 171 | Previous review comment still unresolved: consider returning a `dict` instead of |
| `utilities/shared_disk.py` | 148 | No bounds check — if `idx >= 26`, `chr(ord('a') + idx)` produces non-letter char |

---
*Review generated by pi*

utilities/shared_disk.py
MiriSafra: Done — `_get_shared_disk_devices` now returns `dict[str, str]` mapping destination VM name to device path (e.g., `{"vm1-name": "/dev/vdc", "vm2-name": "/dev/vdb"}`). Caller uses `device_by_vm[vm1_dest_name]` instead of positional unpacking.

utilities/shared_disk.py
MiriSafra: Good catch — fixed. Renamed to `_get_pvc_volume_indexes` which now iterates ALL volumes (including non-PVC like cloudinit) via `enumerate(cnv_vm.vmi.instance.spec.volumes)`, and only records the index for PVC-backed volumes. This ensures the index reflects KubeVirt's actual device letter assignment.

utilities/shared_disk.py
MiriSafra: Addressed together with the first dict comment — `_get_shared_disk_devices` returns `dict[str, str]` keyed by destination VM name.

utilities/shared_disk.py
MiriSafra: Added the exact bounds check you suggested — `ValueError` for `idx > 25`.

utilities/shared_disk.py
```diff
@@ -97,48 +102,123 @@ def _write_marker(ssh_conn: VMSSHConnection, file_path: str, content: str, vm_la
...
-        to match by exact SCSI position instead.
+    Returns:
+        str: Device path (e.g., 0 -> "/dev/vda", 2 -> "/dev/vdc").
+
+    Raises:
+        ValueError: If index exceeds the a-z device range.
+    """
+    if idx > 25:
+        raise ValueError(f"Volume index {idx} exceeds supported device range (a-z)")
+    return f"/dev/vd{chr(ord('a') + idx)}"
```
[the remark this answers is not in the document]
MiriSafra: Good catch — the docstring was misleading. Forklift migrates all disks to virtio bus, so `/dev/vd*` is correct for all current use cases. Fixed the docstring to say "Assumes virtio bus (Forklift default after migration)" instead of claiming multi-bus support.

myakove (commented):
## Code Review

### :bulb: Suggestions (3)

| File | Line | Issue |
|------|------|-------|
| `utilities/shared_disk.py` | 243 | **Duplicate calls:** `resolve_destination_vm_name(vm1_config)` and `resolve_dest |
| `utilities/shared_disk.py` | 150 | Negative index not guarded — `chr(ord('a') + (-1))` produces `/dev/vd\`` which i |
| `utilities/shared_disk.py` | 136 | The volume index from `spec.volumes` is used to derive `/dev/vdX`, but KubeVirt  |

---
*Review generated by pi*

utilities/shared_disk.py
```diff
@@ -151,17 +231,25 @@ def verify_shared_disk_data(
...
+        ocp_admin_client (DynamicClient): OpenShift admin client for destination VM lookup.
 
     Raises:
         AssertionError: If shared disk data verification fails.
         GuestCommandError: If SSH commands fail.
     """
-    vm1_name = prepared_plan["virtual_machines"][0]["name"]
-    vm2_name = prepared_plan["virtual_machines"][1]["name"]
-    shared_disk_device = _get_shared_disk_device(prepared_plan, vm1_name)
+    vm1_config = prepared_plan["virtual_machines"][0]
```
myakove, line 249: **Duplicate calls:** `resolve_destination_vm_name(vm1_config)` and `resolve_destination_vm_name(vm2_config)` are called here AND inside `_get_shared_disk_devices()` — same function, same args, called twice per VM.

Consider computing dest names once here and passing strings to `_get_shared_disk_devices()` instead of config dicts:

```python
vm1_dest_name = resolve_destination_vm_name(vm1_config)
vm2_dest_name = resolve_destination_vm_name(vm2_config)
device_by_vm = _get_shared_disk_devices(ocp_admin_client, vm_namespace, vm1_dest_name, vm2_dest_name)
```

utilities/shared_disk.py
```diff
@@ -97,48 +102,123 @@ def _write_marker(ssh_conn: VMSSHConnection, file_path: str, content: str, vm_la
...
-    order as the source SCSI layout.
+    Args:
+        idx (int): Zero-based volume index.
 
-    Note:
-        Assumes exactly two SCSI controllers with the shared disk alone on the
-        non-boot controller. This matches the current test VM layout. If VMs
-        gain additional controllers with non-shared disks, this logic would need
-        to match by exact SCSI position instead.
+    Returns:
```
myakove: Negative index not guarded — `chr(ord('a') + (-1))` produces `/dev/vd\`` which is nonsensical. Consider:

```suggestion
    if not 0 <= idx <= 25:
        raise ValueError(f"Volume index {idx} exceeds supported device range (a-z)")
```

utilities/shared_disk.py
```diff
@@ -97,48 +102,123 @@ def _write_marker(ssh_conn: VMSSHConnection, file_path: str, content: str, vm_la
...
+    cnv_vm = VirtualMachine(
+        client=ocp_admin_client,
+        name=vm_name,
+        namespace=target_namespace,
+        ensure_exists=True,
+    )
+    if not cnv_vm.vmi:
+        raise ValueError(f"VM '{vm_name}' in '{target_namespace}' has no running VMI")
+
+    pvc_indexes: dict[str, int] = {}
```
myakove: The volume index from `spec.volumes` is used to derive `/dev/vdX`, but KubeVirt assigns device letters from `spec.domain.devices.disks` order. Currently works because Forklift creates both lists in the same order — but this is an implicit assumption.

Consider using `VMI status.volumeStatus[].target` which contains the actual assigned device name (e.g., `"vda"`) — no index math needed and immune to ordering differences.

utilities/shared_disk.py
MiriSafra: Done. `verify_shared_disk_data()` now computes dest names once and passes strings. `_get_shared_disk_devices()` signature changed from config dicts to `vm1_dest_name: str, vm2_dest_name: str`.

utilities/shared_disk.py
MiriSafra: Eliminated entirely — `_index_to_device_path()` and all index arithmetic removed. Device paths now come from `VMI status.volumeStatus[].target` (see comment below), so there's no index-to-letter conversion left to guard.

utilities/shared_disk.py
MiriSafra: Implemented. Renamed `_get_pvc_volume_indexes()` → `_get_pvc_device_targets()` — now reads `volumeStatus[].target` directly from the VMI runtime status. Returns `dict[str, str]` (PVC name → `/dev/{target}`). Added `TimeoutSampler` wait for `volumeStatus` to populate (it's a runtime field that appears after the VM is fully Running). All index math and `_index_to_device_path()` removed.

utilities/shared_disk.py
```diff
@@ -97,48 +104,110 @@ def _write_marker(ssh_conn: VMSSHConnection, file_path: str, content: str, vm_la
...
+            f"VM '{vm_name}' in '{target_namespace}' VMI has no volumeStatus after 300s "
+            f"(phase: {getattr(vmi.instance.status, 'phase', 'unknown')})"
+        ) from None
+
+    pvc_devices: dict[str, str] = {}
+    for vol_status in vmi.instance.status.volumeStatus:
+        pvc_info = getattr(vol_status, "persistentVolumeClaimInfo", None)
+        if pvc_info and vol_status.target:
+            pvc_devices[pvc_info.claimName] = f"/dev/{vol_status.target}"
+    return pvc_devices
```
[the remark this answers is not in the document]
MiriSafra: Fixed — moved PVC-to-device extraction inside the TimeoutSampler loop. Now waits until at least one PVC-backed volume has a populated `target` field before returning, instead of just checking that `volumeStatus` exists. This prevents returning an empty dict during the brief window where volumeStatus is present but targets haven't been assigned yet.

myakove (commented):
## Code Review

### :bulb: Suggestions (6)

| File | Line | Issue |
|------|------|-------|
| `utilities/shared_disk.py` | 131 | **Double `.vmi` property access creates race condition:** Between the `if not cn |
| `utilities/shared_disk.py` | 148 | **TOCTOU gap:** The `TimeoutSampler` loop confirms `sample.status.volumeStatus`  |
| `utilities/shared_disk.py` | 142 | **`from None` suppresses original exception context:** `raise ValueError(...) fr |
| `utilities/shared_disk.py` | 155 | **Silent skip of PVC volumes missing `target` field:** The condition `if pvc_inf |
| `utilities/shared_disk.py` | 136 | **Magic numbers:** `300` and `5` should be named constants for clarity and tunin |
| `utilities/shared_disk.py` | 158 | **No logging after PVC mapping loop:** If `pvc_devices` is unexpectedly empty (e |

---
*Review generated by pi*

utilities/shared_disk.py
```diff
@@ -97,48 +104,110 @@ def _write_marker(ssh_conn: VMSSHConnection, file_path: str, content: str, vm_la
...
+
+    Returns:
+        dict[str, str]: Mapping of PVC claim name to device path,
+            e.g. ``{"pvc-boot": "/dev/vda", "pvc-shared": "/dev/vdc"}``.
+
+    Raises:
+        ValueError: If the VM has no running instance (VMI).
+    """
+    cnv_vm = VirtualMachine(
+        client=ocp_admin_client,
```
myakove, line 136: **Double `.vmi` property access creates race condition:** Between the `if not cnv_vm.vmi` check and the `vmi = cnv_vm.vmi` assignment, the VMI state could change (e.g., VMI terminates between the two calls). Assign once and check:

```suggestion
    vmi = cnv_vm.vmi
    if not vmi:
        raise ValueError(f"VM '{vm_name}' in '{target_namespace}' has no running VMI")
```

utilities/shared_disk.py
```diff
@@ -97,48 +104,110 @@ def _write_marker(ssh_conn: VMSSHConnection, file_path: str, content: str, vm_la
...
+    vmi = cnv_vm.vmi
+    try:
+        for sample in TimeoutSampler(wait_timeout=300, sleep=5, func=lambda: vmi.instance):
+            if getattr(sample.status, "volumeStatus", None):
+                break
+    except TimeoutExpiredError:
+        raise ValueError(
+            f"VM '{vm_name}' in '{target_namespace}' VMI has no volumeStatus after 300s "
+            f"(phase: {getattr(vmi.instance.status, 'phase', 'unknown')})"
+        ) from None
```
myakove: **TOCTOU gap:** The `TimeoutSampler` loop confirms `sample.status.volumeStatus` is populated, then breaks. But the code below re-accesses `vmi.instance.status.volumeStatus` — if `.instance` is a property that hits the API, this is a separate fetch that could return different data (e.g., during VMI restart). Save the passing sample and iterate over it instead:

```python
vmi_instance = None
for sample in TimeoutSampler(wait_timeout=300, sleep=5, func=lambda: vmi.instance):
    if getattr(sample.status, "volumeStatus", None):
        vmi_instance = sample
        break

# Then use vmi_instance below:
pvc_devices: dict[str, str] = {}
for vol_status in vmi_instance.status.volumeStatus:
```

utilities/shared_disk.py
```diff
@@ -97,48 +104,110 @@ def _write_marker(ssh_conn: VMSSHConnection, file_path: str, content: str, vm_la
...
+        namespace=target_namespace,
+        ensure_exists=True,
+    )
+    if not cnv_vm.vmi:
+        raise ValueError(f"VM '{vm_name}' in '{target_namespace}' has no running VMI")
+
+    vmi = cnv_vm.vmi
+    try:
+        for sample in TimeoutSampler(wait_timeout=300, sleep=5, func=lambda: vmi.instance):
+            if getattr(sample.status, "volumeStatus", None):
```
myakove: **`from None` suppresses original exception context:** `raise ValueError(...) from None` discards the `TimeoutExpiredError` traceback, which may carry sampling diagnostics (last sample value, iteration count) useful for CI debugging. Preserve the chain:

```suggestion
    except TimeoutExpiredError as exc:
        raise ValueError(
            f"VM '{vm_name}' in '{target_namespace}' VMI has no volumeStatus after 300s "
            f"(phase: {getattr(vmi.instance.status, 'phase', 'unknown')})"
        ) from exc
```

utilities/shared_disk.py
myakove, line 172: **Silent skip of PVC volumes missing `target` field:** The condition `if pvc_info and vol_status.target:` silently excludes PVC volumes where `target` is `None` (e.g., a disk still being hotplugged). If the shared disk's volume has no `target`, it gets dropped from `pvc_devices` → leading to a misleading "No shared PVC" error downstream that masks the real issue. Raise explicitly when a PVC volume has no target:

```python
for vol_status in vmi_instance.status.volumeStatus:
    pvc_info = getattr(vol_status, "persistentVolumeClaimInfo", None)
    if pvc_info:
        if not vol_status.target:
            raise ValueError(
                f"PVC '{pvc_info.claimName}' on VM '{vm_name}' has no runtime device target"
            )
        pvc_devices[pvc_info.claimName] = f"/dev/{vol_status.target}"
```

utilities/shared_disk.py
```diff
@@ -97,48 +104,110 @@ def _write_marker(ssh_conn: VMSSHConnection, file_path: str, content: str, vm_la
...
+    Raises:
+        ValueError: If the VM has no running instance (VMI).
+    """
+    cnv_vm = VirtualMachine(
+        client=ocp_admin_client,
+        name=vm_name,
+        namespace=target_namespace,
+        ensure_exists=True,
+    )
+    if not cnv_vm.vmi:
```
myakove: **Magic numbers:** `300` and `5` should be named constants for clarity and tuning:

```python
_VMI_VOLUME_STATUS_TIMEOUT = 300  # seconds to wait for VMI volumeStatus population
_VMI_VOLUME_STATUS_POLL_INTERVAL = 5  # seconds between volumeStatus checks
```

utilities/shared_disk.py
```diff
@@ -97,48 +104,110 @@ def _write_marker(ssh_conn: VMSSHConnection, file_path: str, content: str, vm_la
...
+
+    pvc_devices: dict[str, str] = {}
+    for vol_status in vmi.instance.status.volumeStatus:
+        pvc_info = getattr(vol_status, "persistentVolumeClaimInfo", None)
+        if pvc_info and vol_status.target:
+            pvc_devices[pvc_info.claimName] = f"/dev/{vol_status.target}"
+    return pvc_devices
+
+
+def _get_shared_disk_devices(
```
myakove, line 175: **No logging after PVC mapping loop:** If `pvc_devices` is unexpectedly empty (e.g., all volumes lacked `persistentVolumeClaimInfo`), the error only surfaces later in `_get_shared_disk_devices` as "No shared PVC" — making it hard to pinpoint whether the issue was in VM1 or VM2's device resolution. Add:

```python
LOGGER.debug(f"PVC device targets for VM '{vm_name}': {pvc_devices}")
```

utilities/shared_disk.py
MiriSafra: Fixed — `vmi = cnv_vm.vmi` assigned once, then checked. No double access.

utilities/shared_disk.py
MiriSafra: Fixed — the PVC mapping loop now runs inside the TimeoutSampler loop body, iterating over `sample.status.volumeStatus` directly. No separate fetch after the loop.

utilities/shared_disk.py
MiriSafra: Fixed — changed to `from exc` to preserve the `TimeoutExpiredError` traceback. Also uses `sample.status` (from the loop variable) instead of re-fetching `vmi.instance.status`.

utilities/shared_disk.py
MiriSafra: Fixed — missing `target` is now handled explicitly: `pvc_devices.clear(); break` clears partial data and lets TimeoutSampler retry. KubeVirt populates `target` incrementally after VMI startup, so a missing target is a transient state (not a permanent error). If it never populates, the sampler times out with a clear error message. Verified on cluster: VM1 needed one retry (~5s) before targets were populated.

utilities/shared_disk.py
MiriSafra: Fixed — extracted to `_VMI_VOLUME_STATUS_TIMEOUT = 300` and `_VMI_VOLUME_STATUS_POLL_INTERVAL = 5` module-level constants.

utilities/shared_disk.py
MiriSafra: Fixed — added `LOGGER.debug(f"PVC device targets for VM '{vm_name}': {pvc_devices}")` after the mapping loop.

utilities/shared_disk.py
```diff
@@ -97,48 +104,128 @@ def _write_marker(ssh_conn: VMSSHConnection, file_path: str, content: str, vm_la
...
+    try:
+        for sample in TimeoutSampler(
+            wait_timeout=_VMI_VOLUME_STATUS_TIMEOUT,
+            sleep=_VMI_VOLUME_STATUS_POLL_INTERVAL,
+            func=lambda: vmi.instance,
+        ):
+            volume_status = getattr(sample.status, "volumeStatus", None)
+            if not volume_status:
+                continue
+            pvc_devices.clear()
```
myakove, line 154: **`sample` potentially unbound in `except TimeoutExpiredError` handler**

`sample` is only assigned inside the `for sample in TimeoutSampler(...)` loop body. If `TimeoutExpiredError` fires before the first yield (e.g., degenerate timeout/sleep values, or edge-case timing), `sample` is never bound and the `except` block raises `NameError`, masking the real timeout error.

Suggestion — initialize before `try` and guard the access:

```python
sample = None
try:
    for sample in TimeoutSampler(...):
        ...
except TimeoutExpiredError as exc:
    phase = getattr(sample.status, "phase", "unknown") if sample else "no-sample"
    raise ValueError(
        f"VM '{vm_name}' in '{target_namespace}' VMI has no PVC device targets "
        f"after {_VMI_VOLUME_STATUS_TIMEOUT}s (phase: {phase})"
    ) from exc
```

utilities/shared_disk.py
```diff
@@ -97,48 +104,128 @@ def _write_marker(ssh_conn: VMSSHConnection, file_path: str, content: str, vm_la
...
 
-    Note:
-        Assumes exactly two SCSI controllers with the shared disk alone on the
-        non-boot controller. This matches the current test VM layout. If VMs
-        gain additional controllers with non-shared disks, this logic would need
-        to match by exact SCSI position instead.
+    Returns:
+        dict[str, str]: Mapping of PVC claim name to device path,
+            e.g. ``{"pvc-boot": "/dev/vda", "pvc-shared": "/dev/vdc"}``.
+
```
myakove, line 130: **Stale VMI reference captured in lambda**

`vmi = cnv_vm.vmi` is captured once. If the VMI is deleted and recreated during the 300s polling window (e.g., VM restart), this reference becomes stale — `.instance` would return the old/None object, and the polling loop would never succeed.

Consider re-fetching through the VM object each iteration:

```python
func=lambda: cnv_vm.vmi.instance if cnv_vm.vmi else None
```

And add a `None` guard at the top of the loop body.

myakove (commented):
## Code Review

### :bulb: Suggestions (2)

| File | Line | Issue |
|------|------|-------|
| `utilities/shared_disk.py` | 155 | **`sample` potentially unbound in `except TimeoutExpiredError` handler** |
| `utilities/shared_disk.py` | 130 | **Stale VMI reference captured in lambda** |

---
*Review generated by pi*

utilities/shared_disk.py
MiriSafra: Done — initialized `sample = None` before `try` and guarded access in `except` with `if sample` (falls back to `"no-sample"` phase).

utilities/shared_disk.py
MiriSafra: Done — lambda now re-fetches `cnv_vm.vmi` each iteration (`func=lambda: cnv_vm.vmi.instance if cnv_vm.vmi else None`), with `None` guard at the top of the loop body. Also removed the separate `vmi` variable to avoid the stale capture entirely.

myakove (approved):

PR #559, opened 2026-06-16
fix: regenerate manual MAC addresses during cloning with case preservation

krcmarik:
During VM cloning, all NICs get new VMware-generated MAC addresses to prevent MAC conflicts when multiple clones of the same source VM run in parallel (pytest-xdist) - Newly For NICs with manually assigned uppercase MACs, a post-clone ReconfigVM_Task sets the generated MAC to uppercase and restores addressType to "manual", since VMware generates lowercase by default. This preserves the MAC case format needed for static IP preservation testing (MTV-667).


libs/providers/vmware.py
```diff
@@ -1336,21 +1337,24 @@ def clone_vm(
...
+        uppercase_mac_nics: set[str] = set()
         if regenerate_mac:
             source_config = source_vm.config
             if source_config and source_config.hardware and source_config.hardware.device:
                 for device in source_config.hardware.device:
                     if isinstance(device, vim.vm.device.VirtualEthernetCard):
+                        nic_label = device.deviceInfo.label
+                        if device.addressType == "manual":
+                            if any(c.isalpha() and c.isupper() for c in device.macAddress):
+                                uppercase_mac_nics.add(nic_label)
```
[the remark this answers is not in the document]
krcmarik: deviceInfo is never None on VirtualEthernetCard — it's a core property of the VMware managed object. Adding a None guard would be defensive coding against VMware's own API contract, which we don't validate.

libs/providers/vmware.py
```diff
@@ -1336,21 +1337,24 @@ def clone_vm(
...
+                            continue
+                        device.addressType = "generated"
+
```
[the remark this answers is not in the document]
krcmarik: VMware automatically regenerates MAC addresses for all NICs during the clone operation — the continue for manual NICs just avoids adding unnecessary device change specs to the clone request. The post-clone reconfiguration then restores the original manual MAC address and uppercase casing for NICs that had it.

libs/providers/vmware.py
```diff
@@ -1336,21 +1337,24 @@ def clone_vm(
...
                         device_spec.operation = vim.vm.device.VirtualDeviceSpec.Operation.edit
                         device_spec.device = device
-                        device_spec.device.addressType = "generated"
                         device_changes.append(device_spec)
-                        LOGGER.info(
-                            f"Configured MAC regeneration for network device: {
-                                device.deviceInfo.label if device.deviceInfo else 'Unknown'
-                            }",
-                        )
+                        LOGGER.info(f"Configured MAC regeneration for network device: {nic_label}")
```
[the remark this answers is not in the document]
krcmarik: VMware automatically regenerates MACs for all NICs during cloning — the continue skips adding device change specs, not the regeneration itself. All manual NICs get their MACs regenerated by VMware regardless. The post-clone block then restores uppercase-manual NICs to their original MAC and addressType.

utilities/post_migration.py
```diff
@@ -690,7 +690,16 @@ def check_memory(source_vm: dict[str, Any], destination_vm: dict[str, Any]) -> N
...
+    """Get a NIC from a list of NICs by MAC address (case-insensitive).
+
+    Args:
+        nics (list[dict[str, Any]]): List of NIC dictionaries with "macAddress" keys.
+        mac_address (str): The MAC address to search for.
+
+    Returns:
+        dict[str, Any]: The matching NIC dictionary.
+    """
+    return [nic for nic in nics if nic["macAddress"].lower() == mac_address.lower()][0]
```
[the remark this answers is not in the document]
krcmarik: This is a pre-existing pattern on the main branch — the [0] indexing existed before this PR. The only change in this PR is making the comparison case-insensitive. Addressing the IndexError is out of scope for this branch.

libs/providers/vmware.py
```diff
@@ -1475,6 +1479,27 @@ def clone_vm(
...
+        # Only uppercase MACs need reconfiguration since VMware generates lowercase by default.
+        if uppercase_mac_nics and res:
+            reconfig_specs = []
+            for dev in res.config.hardware.device:
+                if (
+                    not isinstance(dev, vim.vm.device.VirtualEthernetCard)
+                    or dev.deviceInfo.label not in uppercase_mac_nics
+                ):
+                    continue
+                dev.macAddress = dev.macAddress.upper()
```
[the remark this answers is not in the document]
krcmarik: macAddress is a guaranteed property on VirtualEthernetCard — it's always present and always a string. Adding None validation would be defensive coding against VMware's own API contract, which we don't validate in this codebase.

utilities/post_migration.py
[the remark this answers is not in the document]
krcmarik: This is a pre-existing pattern on the main branch — the only change in this PR is adding .lower() for case-insensitive comparison. The NIC dicts are built by our own provider code which always populates macAddress. Addressing the IndexError or adding type validation is out of scope for this branch.

libs/providers/vmware.py
```diff
@@ -1475,6 +1479,27 @@ def clone_vm(
         for rdm_config in rdm_disks:
             self.add_rdm_disk_to_vm(vm=res, rdm_type=rdm_config["rdm_type"], enable_cbt=enable_ctk)
 
+        # Reconfigure cloned NICs with uppercase manual MACs back to manual address type.
+        # Only uppercase MACs need reconfiguration since VMware generates lowercase by default.
+        if uppercase_mac_nics and res:
+            reconfig_specs = []
+            for dev in res.config.hardware.device:
+                if (
```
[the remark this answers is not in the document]
krcmarik: The VM object (res) is returned from a successful clone task that completed via wait_task(). A freshly cloned VM always has config.hardware.device populated. The guard on the source VM (source_config and source_config.hardware) exists because the source could theoretically be in a bad state, but a just-cloned VM is fully accessible by definition.

libs/providers/vmware.py
```diff
@@ -1475,6 +1478,27 @@ def clone_vm(
...
+                dev.addressType = "manual"
+                dev_spec = vim.vm.device.VirtualDeviceSpec()
+                dev_spec.operation = vim.vm.device.VirtualDeviceSpec.Operation.edit
+                dev_spec.device = dev
+                reconfig_specs.append(dev_spec)
+                LOGGER.info(f"Set manual MAC for {dev.deviceInfo.label}: {dev.macAddress}")
+            if reconfig_specs:
+                reconfig_task = res.ReconfigVM_Task(spec=vim.vm.ConfigSpec(deviceChange=reconfig_specs))
+                self.wait_task(task=reconfig_task, action_name=f"Reconfiguring manual MACs on {clone_vm_name}")
+
```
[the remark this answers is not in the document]
krcmarik: Skipping — `clone_vm()` was already a long multi-purpose function before this PR. The MAC handling (~15 lines pre-clone + ~20 lines post-clone) is tightly coupled to the cloning flow. Extracting into helpers would fragment the workflow across methods for marginal readability gain. Not a PR-specific issue.

utilities/post_migration.py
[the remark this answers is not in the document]
krcmarik: Skipping — negligible performance impact. The NIC list is typically 1-3 items; recomputing `.lower()` on a 17-character MAC string 1-3 times has zero measurable cost. Not worth a code change.

myakove: **`get_nic_by_mac` raises opaque `IndexError` on no match**

If no NIC matches the MAC address, the list comprehension produces an empty list and `[0]` raises `IndexError` with no context about which MAC was searched or which NICs were available. Per "Fail Fast - Validate Content Not Just Existence" (MUST):

```python
def get_nic_by_mac(nics: list[dict[str, Any]], mac_address: str) -> dict[str, Any]:
    """...

    Raises:
        ValueError: If no NIC matches the MAC address.
    """
    matched = [nic for nic in nics if nic["macAddress"].lower() == mac_address.lower()]
    if not matched:
        raise ValueError(f"No NIC found with MAC address '{mac_address}' in {len(nics)} NICs")
    return matched[0]
```

myakove (commented):
## Code Review

### :bulb: Suggestions (5)

| File | Line | Issue |
|------|------|-------|
| `utilities/post_migration.py` | 702 | **`get_nic_by_mac` raises opaque `IndexError` on no match** |
| `libs/providers/vmware.py` | 1483 | **Post-clone MAC reconfiguration runs even after retry path skipped MAC regenera |
| `libs/providers/vmware.py` | 1346 | **`device.deviceInfo.label` accessed without null check — regression from previo |
| `libs/providers/vmware.py` | 1478 | **Consider extracting post-clone MAC reconfiguration into a helper** |
| `libs/providers/vmware.py` | 1349 | **Simplify uppercase detection** |

---
*Review generated by pi*

libs/providers/vmware.py
```diff
@@ -1475,6 +1478,27 @@ def clone_vm(
         for rdm_config in rdm_disks:
             self.add_rdm_disk_to_vm(vm=res, rdm_type=rdm_config["rdm_type"], enable_cbt=enable_ctk)
 
+        # Reconfigure cloned NICs with uppercase manual MACs back to manual address type.
+        # Only uppercase MACs need reconfiguration since VMware generates lowercase by default.
+        if uppercase_mac_nics and res:
```
myakove: **Post-clone MAC reconfiguration runs even after retry path skipped MAC regeneration**

When the initial clone fails with "in use" (line 1452), the retry removes all NIC device changes to avoid MAC conflicts. However, `uppercase_mac_nics` is still populated from the source VM inspection (line 1349), so this post-clone block still executes — performing an unnecessary `ReconfigVM_Task` API call and logging misleading "Set manual MAC for..." messages.

Suggestion: clear `uppercase_mac_nics` in the retry `except` block:
```python
# In the retry path (around line 1452-1470)
uppercase_mac_nics.clear()
```

libs/providers/vmware.py
```diff
@@ -1336,21 +1337,23 @@ def clone_vm(
             )
             device_changes.extend(disk_device_specs)
 
+        uppercase_mac_nics: set[str] = set()
         if regenerate_mac:
             source_config = source_vm.config
             if source_config and source_config.hardware and source_config.hardware.device:
                 for device in source_config.hardware.device:
                     if isinstance(device, vim.vm.device.VirtualEthernetCard):
+                        nic_label = device.deviceInfo.label
```
myakove, line 1349: **`device.deviceInfo.label` accessed without null check — regression from previous code**

The old code had a defensive check:
```python
device.deviceInfo.label if device.deviceInfo else 'Unknown'
```

The new code unconditionally accesses `device.deviceInfo.label` here and at line 1488 (`dev.deviceInfo.label`). If `deviceInfo` is `None` → `AttributeError`.

Suggestion: keep the defensive check, or validate `device.deviceInfo` is not None before proceeding (fail-fast per project guidelines).

libs/providers/vmware.py
```diff
@@ -1475,6 +1478,27 @@ def clone_vm(
         for rdm_config in rdm_disks:
```
myakove, line 1488: **Consider extracting post-clone MAC reconfiguration into a helper**

`clone_vm` is ~345 lines, well over the 50-line guideline. This 18-line reconfiguration block is self-contained with a clear single responsibility. Extract to `_reconfigure_uppercase_manual_macs(self, vm, clone_vm_name, uppercase_mac_nics)` for readability and testability.

libs/providers/vmware.py
```diff
@@ -1336,21 +1337,23 @@ def clone_vm(
...
+        uppercase_mac_nics: set[str] = set()
         if regenerate_mac:
             source_config = source_vm.config
             if source_config and source_config.hardware and source_config.hardware.device:
                 for device in source_config.hardware.device:
                     if isinstance(device, vim.vm.device.VirtualEthernetCard):
+                        nic_label = device.deviceInfo.label
+                        if device.addressType == "manual":
+                            if any(c.isalpha() and c.isupper() for c in device.macAddress):
+                                uppercase_mac_nics.add(nic_label)
```
myakove, line 1355: **Simplify uppercase detection**

`any(c.isalpha() and c.isupper() for c in device.macAddress)` iterates character-by-character. A simpler equivalent:

```suggestion
                            if device.macAddress != device.macAddress.lower():
```

Directly expresses the intent ("MAC contains uppercase characters") without character-level iteration.

utilities/post_migration.py
krcmarik: Good catch. Replaced the bare [0] index with explicit validation — now raises ValueError with the searched MAC and NIC count for easier debugging.

libs/providers/vmware.py
krcmarik: Right, added uppercase_mac_nics.clear() in the retry block so the post-clone reconfiguration is skipped when MAC regeneration was disabled during retry.

libs/providers/vmware.py
krcmarik: Added deviceInfo null checks before accessing .label at both locations (source VM scan and post-clone reconfig). Raises ValueError with VM name context instead of falling back to 'Unknown'.

libs/providers/vmware.py
krcmarik: Extracted to _reconfigure_uppercase_manual_macs() with proper docstring and deviceInfo validation.

libs/providers/vmware.py
krcmarik: Done. Using device.macAddress != device.macAddress.lower() — cleaner and directly expresses the intent.

redhat-qe-bot1:
**Clean rebase detected** — no code changes compared to previous head (`f123ee7`).
The following labels were preserved: `commented-krcmarik`, `lgtm-coderabbitai[bot]`.

libs/providers/vmware.py
```diff
@@ -1475,8 +1481,43 @@ def clone_vm(
...
+        """Reconfigure cloned NICs with uppercase manual MACs back to manual address type.
+
+        VMware generates lowercase MACs by default, so only uppercase MACs need
+        reconfiguration after cloning.
+
+        Args:
+            vm (Any): The cloned VM object to reconfigure.
+            clone_vm_name (str): Name of the cloned VM (used for logging and error messages).
+            uppercase_mac_nics (set[str]): Set of NIC labels whose MACs need uppercasing.
+        """
```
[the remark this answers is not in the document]
krcmarik: Per project docstring rules: 'Returns: include only when returning a meaningful non-None value (omit for void functions).' This is a void function (-> None), so Returns is correctly omitted. Raises could be added but the ValueError is a defensive guard for an edge case that shouldn't occur in practice (pyVmomi always populates deviceInfo on VirtualEthernetCard).

libs/providers/vmware.py
```diff
@@ -1336,21 +1337,25 @@ def clone_vm(
...
             device_changes.extend(disk_device_specs)
 
+        uppercase_mac_nics: set[str] = set()
         if regenerate_mac:
             source_config = source_vm.config
             if source_config and source_config.hardware and source_config.hardware.device:
                 for device in source_config.hardware.device:
                     if isinstance(device, vim.vm.device.VirtualEthernetCard):
+                        if not device.deviceInfo:
+                            raise ValueError(f"Device has no deviceInfo in VM '{source_vm_name}' config")
```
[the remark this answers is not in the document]
krcmarik: ValueError is intentional here. The deviceInfo check is a data integrity validation, not a clone operation failure. VmCloneError is for clone task failures (e.g., resource conflicts, vCenter errors). A missing deviceInfo on a VirtualEthernetCard would indicate a pyVmomi/vCenter data anomaly — a fundamentally different error category. Per project guidelines: 'Missing resource → ValueError'.

libs/providers/vmware.py
```diff
@@ -1475,8 +1481,43 @@ def clone_vm(
...
+            dev.macAddress = dev.macAddress.upper()
+            dev.addressType = "manual"
+            dev_spec = vim.vm.device.VirtualDeviceSpec()
+            dev_spec.operation = vim.vm.device.VirtualDeviceSpec.Operation.edit
+            dev_spec.device = dev
+            reconfig_specs.append(dev_spec)
+            LOGGER.info(f"Set manual MAC for {dev.deviceInfo.label}: {dev.macAddress}")
+        if reconfig_specs:
+            reconfig_task = vm.ReconfigVM_Task(spec=vim.vm.ConfigSpec(deviceChange=reconfig_specs))
+            self.wait_task(task=reconfig_task, action_name=f"Reconfiguring manual MACs on {clone_vm_name}")
```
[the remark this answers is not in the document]
krcmarik: The default 60s timeout is sufficient for a simple NIC MAC address reconfiguration (changing addressType and macAddress on a few NICs). The 120s timeout on add_rdm_disk_to_vm is for a heavier operation (adding physical disk mappings). Different operations warrant different timeouts based on their complexity, not a blanket value.

myakove (commented):
## Code Review

### :bulb: Suggestions (5)

| File | Line | Issue |
|------|------|-------|
| `libs/providers/vmware.py` | 1350 | **Lowercase manual MACs silently lose their address during cloning** |
| `libs/providers/vmware.py` | 1491 | **`vm: Any` — use `vim.VirtualMachine` instead** |
| `libs/providers/vmware.py` | 1484 | **Silent skip when `res` is `None` — violates Fail Fast** |
| `libs/providers/vmware.py` | 1502 | **`vm.config.hardware.device` accessed without null checks** |
| `libs/providers/vmware.py` | 1340 | **`uppercase_mac_nics` initialized outside `if regenerate_mac` scope** |

---
*Review generated by pi*

libs/providers/vmware.py
```diff
@@ -1336,21 +1337,25 @@ def clone_vm(
...
         if regenerate_mac:
             source_config = source_vm.config
             if source_config and source_config.hardware and source_config.hardware.device:
                 for device in source_config.hardware.device:
                     if isinstance(device, vim.vm.device.VirtualEthernetCard):
+                        if not device.deviceInfo:
+                            raise ValueError(f"Device has no deviceInfo in VM '{source_vm_name}' config")
+                        nic_label = device.deviceInfo.label
+                        if device.addressType == "manual":
+                            if device.macAddress != device.macAddress.lower():
```
myakove, line 1354: **Lowercase manual MACs silently lose their address during cloning**

All manual NICs get `addressType = "generated"` (line 1353), but only NICs with **uppercase** MACs are tracked in `uppercase_mac_nics` for post-clone restoration. A NIC with `addressType="manual"` and a lowercase MAC (e.g., `00:50:56:aa:bb:cc`) will:
1. Have its `addressType` changed to `generated` → VMware assigns a new random MAC
2. NOT be added to `uppercase_mac_nics` → NOT restored to `manual` post-clone
3. Lose its original MAC address

The docstring says "only uppercase MACs need reconfiguration" — is this intentional VMware behavior (lowercase MACs are always auto-generated and never manually assigned)? If so, please add a code comment explaining why.

If ALL manual MACs should be preserved regardless of case, track all manual NICs:
```python
if device.addressType == "manual":
    manual_mac_nics[nic_label] = device.macAddress  # preserve original MAC
```

libs/providers/vmware.py
```diff
@@ -1475,8 +1481,43 @@ def clone_vm(
...
             self.add_rdm_disk_to_vm(vm=res, rdm_type=rdm_config["rdm_type"], enable_cbt=enable_ctk)
 
+        if uppercase_mac_nics and res:
+            self._reconfigure_uppercase_manual_macs(
+                vm=res, clone_vm_name=clone_vm_name, uppercase_mac_nics=uppercase_mac_nics
+            )
+
         return res
 
+    def _reconfigure_uppercase_manual_macs(self, vm: Any, clone_vm_name: str, uppercase_mac_nics: set[str]) -> None:
```
myakove: **`vm: Any` — use `vim.VirtualMachine` instead**

Per project guidelines: "Using `Any` must be justified — don't use it as a shortcut when the actual type is known." The codebase consistently uses `vim.VirtualMachine` for VM parameters (lines 334, 388, 472, 743, 777, etc.).

```suggestion
    def _reconfigure_uppercase_manual_macs(self, vm: vim.VirtualMachine, clone_vm_name: str, uppercase_mac_nics: set[str]) -> None:
```

libs/providers/vmware.py
```diff
@@ -1475,8 +1481,43 @@ def clone_vm(
         for rdm_config in rdm_disks:
             self.add_rdm_disk_to_vm(vm=res, rdm_type=rdm_config["rdm_type"], enable_cbt=enable_ctk)
 
+        if uppercase_mac_nics and res:
```
myakove: **Silent skip when `res` is `None` — violates Fail Fast**

`if uppercase_mac_nics and res:` silently skips MAC reconfiguration if `res` is `None`. If `wait_task` succeeded without raising, `res` should be a valid VM. A `None` result indicates an unexpected state that should fail immediately rather than silently continuing.

**Fix:** Validate `res` explicitly after `wait_task` returns:
```python
if not res:
    raise VmCloneError(f"Clone task for '{clone_vm_name}' returned no VM object")

if uppercase_mac_nics:
    self._reconfigure_uppercase_manual_macs(...)
```

Note: the pre-existing `if res and self.fixture_store:` (line 1477) has the same pattern but is outside this PR's scope.

libs/providers/vmware.py
```diff
@@ -1475,8 +1481,43 @@ def clone_vm(
...
+
+        VMware generates lowercase MACs by default, so only uppercase MACs need
+        reconfiguration after cloning.
+
+        Args:
+            vm (Any): The cloned VM object to reconfigure.
+            clone_vm_name (str): Name of the cloned VM (used for logging and error messages).
+            uppercase_mac_nics (set[str]): Set of NIC labels whose MACs need uppercasing.
+        """
+        reconfig_specs = []
```
myakove, line 1521: **`vm.config.hardware.device` accessed without null checks**

The caller (`clone_vm`, line 1343) defensively checks `if source_config and source_config.hardware and source_config.hardware.device`. This helper accesses the same chain on the cloned VM without guards — if any is `None`, it raises `AttributeError` with no context.

**Fix:** Add fail-fast validation:
```python
def _reconfigure_uppercase_manual_macs(self, vm, clone_vm_name, uppercase_mac_nics):
    if not (vm.config and vm.config.hardware and vm.config.hardware.device):
        raise ValueError(f"Cloned VM '{clone_vm_name}' has no hardware device configuration")
    ...
```

libs/providers/vmware.py
```diff
@@ -1336,21 +1337,25 @@ def clone_vm(
             )
             device_changes.extend(disk_device_specs)
 
+        uppercase_mac_nics: set[str] = set()
```
myakove, line 1341: **`uppercase_mac_nics` initialized outside `if regenerate_mac` scope**

This variable is initialized as an empty set here (line 1340), before the `if regenerate_mac:` guard. It exists even when `regenerate_mac=False`, and the post-clone check on line 1484 (`if uppercase_mac_nics and res:`) evaluates it regardless — always `False` for the non-regenerate path, but the scope is misleading.

Suggestion: Either move initialization inside `if regenerate_mac:` and guard the post-clone call with `if regenerate_mac and uppercase_mac_nics:`, or add a comment explaining the cross-block usage.

libs/providers/vmware.py
krcmarik: This is intentional. The purpose is MAC regeneration (to avoid conflicts on cloned VMs), not MAC preservation. All NICs get `addressType="generated"` so VMware assigns new MACs for the clone.

The post-clone reconfiguration only targets uppercase manual MACs because VMware always generates lowercase MACs. A manual NIC with a lowercase MAC gets a new generated lowercase MAC — the case and format are already correct. Only uppercase manual MACs need the post-clone fixup to restore the case (`upper()`) and `addressType="manual"`, since VMware's generated MAC would otherwise be lowercase.

libs/providers/vmware.py
krcmarik: Fixed — changed to `vim.VirtualMachine` and updated the docstring accordingly.

libs/providers/vmware.py
krcmarik: Fixed — added explicit `if not res: raise VmCloneError(...)` validation right after the teardown registration. Placed it before the RDM disk loop as well, since `add_rdm_disk_to_vm` also uses `res`.

libs/providers/vmware.py
krcmarik: Fixed — added fail-fast validation at the top of the method, consistent with the caller's defensive pattern.

libs/providers/vmware.py
krcmarik: Fixed — moved `uppercase_mac_nics` initialization inside the `if regenerate_mac:` block, and updated the post-clone guard to `if regenerate_mac and uppercase_mac_nics:`.

redhat-qe-bot1:
**Clean rebase detected** — no code changes compared to previous head (`718b2f7`).
The following labels were preserved: `commented-krcmarik`, `lgtm-coderabbitai[bot]`.

libs/providers/vmware.py
```diff
@@ -1471,12 +1477,55 @@ def clone_vm(
...
+            raise ValueError(f"Cloned VM '{clone_vm_name}' has no hardware device configuration")
+
+        reconfig_specs = []
+        for dev in vm.config.hardware.device:
+            if not isinstance(dev, vim.vm.device.VirtualEthernetCard):
+                continue
+            if not dev.deviceInfo:
+                raise ValueError(f"Device has no deviceInfo in cloned VM '{clone_vm_name}'")
+            if dev.deviceInfo.label not in uppercase_mac_nics:
+                continue
```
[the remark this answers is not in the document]
krcmarik: VMware NIC labels (`Network adapter 1`, `Network adapter 2`, etc.) are assigned by vSphere and preserved across cloning — the cloned VM gets an identical hardware layout with the same labels. Labels are not user-editable display strings in this context; they're stable identifiers within a VM's device configuration. A label mismatch between source and clone would indicate a VMware platform bug, not a normal scenario. The label-based correlation is reliable for this use case and consistent with how other parts of the codebase identify NICs.

redhat-qe-bot2:
**Clean rebase detected** — no code changes compared to previous head (`2fac452`).
The following labels were preserved: `verified`, `commented-krcmarik`, `lgtm-coderabbitai[bot]`, `commented-qodo-code-review[bot]`.

krcmarik:
@coderabbitai review

krcmarik:
@coderabbitai full review

libs/providers/vmware.py
```diff
@@ -1471,12 +1477,55 @@ def clone_vm(
...
+            dev.macAddress = dev.macAddress.upper()
+            dev.addressType = "manual"
+            dev_spec = vim.vm.device.VirtualDeviceSpec()
+            dev_spec.operation = vim.vm.device.VirtualDeviceSpec.Operation.edit
+            dev_spec.device = dev
+            reconfig_specs.append(dev_spec)
+            LOGGER.info(f"Set manual MAC for {dev.deviceInfo.label}: {dev.macAddress}")
+        if reconfig_specs:
+            reconfig_task = vm.ReconfigVM_Task(spec=vim.vm.ConfigSpec(deviceChange=reconfig_specs))
+            self.wait_task(task=reconfig_task, action_name=f"Reconfiguring manual MACs on {clone_vm_name}")
```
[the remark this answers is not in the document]
krcmarik: Skipping — this scenario (tracked NIC missing from cloned VM hardware) is not realistic. VMware NIC labels are stable across cloning, so a NIC present in the source VM will always be present in the clone. Adding a missing-NIC check here would be overengineering for a case that doesn't occur in practice.

libs/providers/vmware.py
```diff
@@ -1337,20 +1338,24 @@ def clone_vm(
...
+            uppercase_mac_nics: set[str] = set()
             source_config = source_vm.config
             if source_config and source_config.hardware and source_config.hardware.device:
                 for device in source_config.hardware.device:
                     if isinstance(device, vim.vm.device.VirtualEthernetCard):
+                        if not device.deviceInfo:
+                            raise ValueError(f"Device has no deviceInfo in VM '{source_vm_name}' config")
+                        nic_label = device.deviceInfo.label
+                        if device.addressType == "manual":
+                            if device.macAddress != device.macAddress.lower():
```
myakove, line 1354: ⚠️ **[PREV-UNRESOLVED] Lowercase manual MACs — code comment still missing**

Your explanation makes sense — VMware always generates lowercase MACs, so only uppercase manual MACs need post-clone restoration. However, the original review asked for a **code comment** explaining this non-obvious behavior, and none was added.

The logic "only track uppercase manual MACs" is counter-intuitive without context. Please add a comment like:
```python
# VMware generates lowercase MACs by default. Only uppercase manual MACs
# need post-clone restoration — lowercase manual MACs get a new generated
# MAC that already matches the expected case/format.
if device.macAddress != device.macAddress.lower():
    uppercase_mac_nics.add(nic_label)
```

libs/providers/vmware.py
```diff
@@ -1337,20 +1338,24 @@ def clone_vm(
             device_changes.extend(disk_device_specs)
 
         if regenerate_mac:
+            uppercase_mac_nics: set[str] = set()
```
myakove: ⚠️ **Fragile `uppercase_mac_nics` variable scoping**

`uppercase_mac_nics` is initialized inside `if regenerate_mac:` (here) but referenced at lines 1456 (`uppercase_mac_nics.clear()`) and 1487 (`if regenerate_mac and uppercase_mac_nics:`) outside this block. Python short-circuit evaluation makes this safe today, but a future refactor separating the `regenerate_mac` guard from either reference would cause a `NameError` at runtime.

Suggestion: initialize before the `if regenerate_mac:` block:
```python
uppercase_mac_nics: set[str] = set()
if regenerate_mac:
    source_config = source_vm.config
    ...
```

libs/providers/vmware.py
```diff
@@ -1471,12 +1477,55 @@ def clone_vm(
         if res and self.fixture_store:
             self.fixture_store["teardown"].setdefault(self.type, []).append({"name": clone_vm_name})
```
myakove, line 1482: ⚠️ **Clone VM not registered for teardown when `res` is falsy — resource leak risk**

Lines 1478-1479 register `clone_vm_name` for teardown only when `res` is truthy. Line 1481 raises `VmCloneError` when `res` is falsy. If `wait_task` succeeds (task state = success) but `task.info.result` is `None`, the clone VM may exist on vSphere but never gets registered for cleanup — the raised exception prevents the caller from getting the VM object, so the VM leaks.

Suggestion: register for teardown unconditionally BEFORE the validity check:
```python
if self.fixture_store:
    self.fixture_store["teardown"].setdefault(self.type, []).append({"name": clone_vm_name})

if not res:
    raise VmCloneError(f"Clone task for '{clone_vm_name}' returned no VM object")
```

utilities/post_migration.py
```diff
@@ -777,7 +777,22 @@ def check_memory(source_vm: dict[str, Any], destination_vm: dict[str, Any]) -> N
...
+
+    Returns:
+        dict[str, Any]: The matching NIC dictionary.
+
+    Raises:
+        ValueError: If no NIC matches the MAC address.
+    """
+    matched = [nic for nic in nics if nic["macAddress"].lower() == mac_address.lower()]
+    if not matched:
+        raise ValueError(f"No NIC found with MAC address '{mac_address}' in {len(nics)} NICs")
```
myakove: 💡 **Error message could include available MACs for easier debugging**

The `ValueError` reports the count of NICs but not their actual MAC addresses. When debugging MAC mismatches across providers, knowing what MACs are available is essential.

Suggestion:
```python
raise ValueError(
    f"No NIC found with MAC address '{mac_address}' in {len(nics)} NICs "
    f"(available: {[n['macAddress'] for n in nics]})"
)
```

myakove (commented):
## Code Review

Found **4** issue(s) in this PR:

### :bulb: Suggestions (4)

| File | Line | Issue |
|------|------|-------|
| `libs/providers/vmware.py` | 1350 | ⚠️ **[PREV-UNRESOLVED] Lowercase manual MACs — code comment still missing** |
| `libs/providers/vmware.py` | 1341 | ⚠️ **Fragile `uppercase_mac_nics` variable scoping** |
| `libs/providers/vmware.py` | 1478 | ⚠️ **Clone VM not registered for teardown when `res` is falsy — resource leak ri |
| `utilities/post_migration.py` | 794 | 💡 **Error message could include available MACs for easier debugging** |

---
*Review generated by pi*

libs/providers/vmware.py
krcmarik: Addressed in the follow-up thread — code comment has been added explaining the uppercase-only MAC tracking logic.

libs/providers/vmware.py
krcmarik: Fixed. Added the code comment explaining why only uppercase manual MACs need post-clone restoration:
```python
# VMware generates lowercase MACs by default. Only uppercase manual MACs
# need post-clone restoration — lowercase manual MACs get a new generated
# MAC that already matches the expected case/format.
```

libs/providers/vmware.py
krcmarik: Fixed. Moved `uppercase_mac_nics: set[str] = set()` initialization before the `if regenerate_mac:` block.

libs/providers/vmware.py
krcmarik: Fixed. Removed `res and` from the condition — clone VM is now registered for teardown unconditionally before the validity check.

utilities/post_migration.py
krcmarik: Fixed. Error message now includes available MAC addresses:
```python
f"(available: {[n['macAddress'] for n in nics]})"
```

libs/providers/vmware.py
```diff
@@ -1468,15 +1477,58 @@ def clone_vm(
             else:
                 raise
 
-        if res and self.fixture_store:
+        if self.fixture_store:
             self.fixture_store["teardown"].setdefault(self.type, []).append({"name": clone_vm_name})
```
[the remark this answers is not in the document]
krcmarik: Skipped. The `if self.fixture_store:` truthiness check is pre-existing (was `if res and self.fixture_store:` before this PR). `fixture_store` is never an empty dict in practice — it's always populated with `session_uuid`, `teardown`, etc. when provided. Low value change.

libs/providers/vmware.py
```diff
@@ -1468,15 +1477,58 @@ def clone_vm(
...
 
+        if not res:
+            raise VmCloneError(f"Clone task for '{clone_vm_name}' returned no VM object")
```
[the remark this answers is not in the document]
krcmarik: Skipped. This directly contradicts a resolved human review comment (myakove) which specifically requested unconditional teardown registration BEFORE the validity check to prevent resource leaks on vSphere. A leaked VM is worse than a teardown attempt on a non-existent VM — teardown should handle missing resources gracefully.

libs/providers/vmware.py
```diff
@@ -1468,15 +1477,58 @@ def clone_vm(
...
+            if dev.deviceInfo.label not in uppercase_mac_nics:
+                continue
+            dev.macAddress = dev.macAddress.upper()
+            dev.addressType = "manual"
+            dev_spec = vim.vm.device.VirtualDeviceSpec()
+            dev_spec.operation = vim.vm.device.VirtualDeviceSpec.Operation.edit
+            dev_spec.device = dev
+            reconfig_specs.append(dev_spec)
+            LOGGER.info(f"Set manual MAC for {dev.deviceInfo.label}: {dev.macAddress}")
+        if reconfig_specs:
```
myakove, line 1542: 💡 **Silent no-op when uppercase NIC labels not found in cloned VM**

If `uppercase_mac_nics` has labels but no matching NICs are found in the cloned VM (e.g., a NIC was dropped during cloning), `reconfig_specs` stays empty and the method returns silently — no warning, no error.

Suggestion: log a warning when expected NICs aren't matched:
```python
if reconfig_specs:
    reconfig_task = vm.ReconfigVM_Task(spec=vim.vm.ConfigSpec(deviceChange=reconfig_specs))
    self.wait_task(task=reconfig_task, action_name=f"Reconfiguring manual MACs on {clone_vm_name}")

unmatched = uppercase_mac_nics - {spec.device.deviceInfo.label for spec in reconfig_specs}
if unmatched:
    LOGGER.warning(f"NICs not found in cloned VM '{clone_vm_name}' for MAC restoration: {unmatched}")
```

myakove (commented):
## Code Review

Found **1** issue(s) in this PR:

### :bulb: Suggestions (1)

| File | Line | Issue |
|------|------|-------|
| `libs/providers/vmware.py` | 1528 | 💡 **Silent no-op when uppercase NIC labels not found in cloned VM** |

---
*Review generated by pi*

libs/providers/vmware.py
krcmarik: Added the warning for unmatched NICs. Worth noting that this scenario (a NIC disappearing during `CloneVM_Task`) would indicate a vSphere platform bug rather than an MTV issue — the `uppercase_mac_nics` set is built from the source VM's devices right before cloning. Still, the warning is cheap and could help debug unexpected clone behavior.

libs/providers/vmware.py
```diff
@@ -1468,15 +1477,62 @@ def clone_vm(
...
+            reconfig_task = vm.ReconfigVM_Task(spec=vim.vm.ConfigSpec(deviceChange=reconfig_specs))
+            self.wait_task(task=reconfig_task, action_name=f"Reconfiguring manual MACs on {clone_vm_name}")
+
```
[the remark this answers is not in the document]
krcmarik: Inlined — removed the intermediate variable, passing the task directly to wait_task().

utilities/post_migration.py
```diff
@@ -777,7 +777,25 @@ def check_memory(source_vm: dict[str, Any], destination_vm: dict[str, Any]) -> N
...
+    Raises:
+        ValueError: If no NIC matches the MAC address.
+    """
+    matched = [nic for nic in nics if nic["macAddress"].lower() == mac_address.lower()]
+    if not matched:
+        raise ValueError(
+            f"No NIC found with MAC address '{mac_address}' in {len(nics)} NICs "
+            f"(available: {[n['macAddress'] for n in nics]})"
+        )
+    return matched[0]
```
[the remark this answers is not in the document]
krcmarik: MAC addresses are unique hardware identifiers per NIC — two NICs with case-only-different MACs would be a hypervisor bug, not something the test suite should defend against. Adding this check would be over-engineering for an impossible scenario.

krcmarik:
@coderabbitai[bot]

The following review comments were reviewed and a decision was made:

### `libs/providers/vmware.py:1505` (outside diff) — MEDIUM: Document the helper’s `ValueError` paths.
> Addressed: Added Raises section documenting both ValueError paths (no hardware device configuration, missing deviceInfo).

myakove (commented):
## Code Review

Found **1** issue(s) in this PR:

### :bulb: Suggestions (1)

| File | Line | Issue |
|------|------|-------|
| `libs/providers/vmware.py` | 1540 | **Unmatched NICs should raise instead of warning** |

---
*Review generated by pi*

libs/providers/vmware.py
```diff
@@ -1468,15 +1477,68 @@ def clone_vm(
...
+            LOGGER.info(f"Set manual MAC for {dev.deviceInfo.label}: {dev.macAddress}")
+        if reconfig_specs:
+            self.wait_task(
+                task=vm.ReconfigVM_Task(spec=vim.vm.ConfigSpec(deviceChange=reconfig_specs)),
+                action_name=f"Reconfiguring manual MACs on {clone_vm_name}",
+            )
+
+        unmatched = uppercase_mac_nics - {spec.device.deviceInfo.label for spec in reconfig_specs}
+        if unmatched:
+            LOGGER.warning(f"NICs not found in cloned VM '{clone_vm_name}' for MAC restoration: {unmatched}")
```
myakove: **Unmatched NICs should raise instead of warning**

If NICs from `uppercase_mac_nics` aren't found in the cloned VM, the MAC case restoration was not applied — but the function returns successfully. Even if a NIC disappearing during `CloneVM_Task` is a vSphere platform bug, the test should still fail fast rather than proceed with incorrect MAC configuration. A downstream test relying on uppercase MAC preservation will fail later with an opaque mismatch error.

Per "No Fallbacks, No Silent Recovery (MUST)": _"LOGGER.warning(...) used to paper over a failure instead of raising"_

Suggestion:
```python
if unmatched:
    raise VmCloneError(
        f"NICs not found in cloned VM '{clone_vm_name}' for MAC restoration: {unmatched}"
    )
```

krcmarik: Fixed: Changed `LOGGER.warning` to `raise VmCloneError` for unmatched NICs, as suggested. Also updated the docstring `Raises` section to document the new exception.

libs/providers/vmware.py
```diff
@@ -1468,15 +1477,70 @@ def clone_vm(
...
+            LOGGER.info(f"Set manual MAC for {dev.deviceInfo.label}: {dev.macAddress}")
+        if reconfig_specs:
+            self.wait_task(
+                task=vm.ReconfigVM_Task(spec=vim.vm.ConfigSpec(deviceChange=reconfig_specs)),
+                action_name=f"Reconfiguring manual MACs on {clone_vm_name}",
+            )
+
+        unmatched = uppercase_mac_nics - {spec.device.deviceInfo.label for spec in reconfig_specs}
+        if unmatched:
+            raise VmCloneError(f"NICs not found in cloned VM '{clone_vm_name}' for MAC restoration: {unmatched}")
```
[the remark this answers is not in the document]
krcmarik: Skipping: Low impact — non-deterministic set ordering in an exception message doesn't affect functionality or debugging. This is not a real bug; the error provides all needed context (VM name + unmatched NIC labels) regardless of label ordering.

libs/providers/vmware.py
```diff
@@ -1468,15 +1477,70 @@ def clone_vm(
...
+            dev_spec.device = dev
+            reconfig_specs.append(dev_spec)
+            LOGGER.info(f"Set manual MAC for {dev.deviceInfo.label}: {dev.macAddress}")
+        if reconfig_specs:
+            self.wait_task(
+                task=vm.ReconfigVM_Task(spec=vim.vm.ConfigSpec(deviceChange=reconfig_specs)),
+                action_name=f"Reconfiguring manual MACs on {clone_vm_name}",
+            )
+
+        unmatched = uppercase_mac_nics - {spec.device.deviceInfo.label for spec in reconfig_specs}
```
myakove: **Validation after mutation: unmatched NICs check runs AFTER `ReconfigVM_Task`**

If some expected NICs are found (and reconfigured) but others are missing, `ReconfigVM_Task` has already been submitted and executed by line 1535. The `VmCloneError` raised here leaves the VM in a partially reconfigured state — some NICs set to `manual`/uppercase, error reported for the rest.

Suggestion: move the unmatched check **before** `wait_task`:
```python
unmatched = uppercase_mac_nics - {spec.device.deviceInfo.label for spec in reconfig_specs}
if unmatched:
    raise VmCloneError(
        f"NICs not found in cloned VM '{clone_vm_name}' for MAC restoration: {unmatched}"
    )
if reconfig_specs:
    self.wait_task(
        task=vm.ReconfigVM_Task(spec=vim.vm.ConfigSpec(deviceChange=reconfig_specs)),
        action_name=f"Reconfiguring manual MACs on {clone_vm_name}",
    )
```

myakove (commented):
## Code Review

### :bulb: Suggestions (5)

| File | Line | Issue |
|------|------|-------|
| `libs/providers/vmware.py` | 1540 | **Validation after mutation: unmatched NICs check runs AFTER `ReconfigVM_Task`** |
| `libs/providers/vmware.py` | 1340 | **`clone_vm` is 338 lines — far exceeding the 50-line guideline** |
| `libs/providers/vmware.py` | 1535 | **`wait_task` always raises `VmCloneError` regardless of operation type** |
| `libs/providers/vmware.py` | 1346 | **`clone_vm` docstring `Raises` section is incomplete** |
| `libs/providers/vmware.py` | 1542 | 💡 **Set comprehension built from mutated device objects is fragile** |

---
*Review generated by pi*

libs/providers/vmware.py
```diff
@@ -1336,21 +1337,28 @@ def clone_vm(
             )
             device_changes.extend(disk_device_specs)
 
+        uppercase_mac_nics: set[str] = set()
```
myakove, line 1341: **`clone_vm` is 338 lines — far exceeding the 50-line guideline**

This PR adds ~20 more lines to an already oversized function. The MAC regeneration block (lines 1340–1367) and CTK/CBT configuration block (lines 1370–1445) are strong extraction candidates into private helpers like `_prepare_mac_regeneration_specs()` and `_prepare_ctk_config()`.

Per "Function Size and Responsibilities (SHOULD)": _"Keep functions under 50 lines when possible. Longer functions need clear section comments."_

This is pre-existing tech debt, but each addition increases the cost of future refactoring.

libs/providers/vmware.py
```diff
@@ -1468,15 +1477,70 @@ def clone_vm(
...
+                continue
+            dev.macAddress = dev.macAddress.upper()
+            dev.addressType = "manual"
+            dev_spec = vim.vm.device.VirtualDeviceSpec()
+            dev_spec.operation = vim.vm.device.VirtualDeviceSpec.Operation.edit
+            dev_spec.device = dev
+            reconfig_specs.append(dev_spec)
+            LOGGER.info(f"Set manual MAC for {dev.deviceInfo.label}: {dev.macAddress}")
+        if reconfig_specs:
+            self.wait_task(
```
myakove, line 1543: **`wait_task` always raises `VmCloneError` regardless of operation type**

`_reconfigure_uppercase_manual_macs` calls `self.wait_task()` for a `ReconfigVM_Task`. If the task fails, `wait_task` (line 306) raises `VmCloneError` — semantically wrong since this is a reconfiguration failure, not a clone failure. Callers catching `VmCloneError` would incorrectly treat a reconfiguration failure as a clone failure.

Suggestion: either generalize `wait_task` to accept a configurable exception type, or wrap the `wait_task` call with a try/except to re-raise as a more appropriate exception (e.g., `VmReconfigError` or at minimum document this behavior).

libs/providers/vmware.py
```diff
@@ -1336,21 +1337,28 @@ def clone_vm(
             )
             device_changes.extend(disk_device_specs)
 
+        uppercase_mac_nics: set[str] = set()
         if regenerate_mac:
             source_config = source_vm.config
             if source_config and source_config.hardware and source_config.hardware.device:
                 for device in source_config.hardware.device:
                     if isinstance(device, vim.vm.device.VirtualEthernetCard):
+                        if not device.deviceInfo:
```
myakove, line 1347: **`clone_vm` docstring `Raises` section is incomplete**

The function now raises `ValueError` here (for missing `deviceInfo` during MAC regeneration) but this isn't documented in `clone_vm`'s `Raises` section (line 1199). Per "Docstring Format (MUST)": _"Raises: include only when exceptions are actually raised."_

Suggestion: add to the `Raises` section:
```python
ValueError: If a network device has no deviceInfo in the source VM config.
```

libs/providers/vmware.py
myakove: 💡 **Set comprehension built from mutated device objects is fragile**

The set `{spec.device.deviceInfo.label for spec in reconfig_specs}` accesses `spec.device` — the same object mutated in-place earlier (`macAddress` uppercased, `addressType` set to `manual`). This works today because `deviceInfo.label` isn't mutated, but creates a fragile coupling.

Suggestion: collect matched labels in a separate set during the loop:
```python
matched_labels: set[str] = set()
for dev in vm.config.hardware.device:
    ...
    matched_labels.add(dev.deviceInfo.label)
    reconfig_specs.append(dev_spec)
...
unmatched = uppercase_mac_nics - matched_labels
```

libs/providers/vmware.py
krcmarik: Fixed: Moved the unmatched NICs check before `wait_task` so we fail-fast before the `ReconfigVM_Task` API call.

**Note:** Combined with your suggestion from the set-comprehension comment — the unmatched computation uses `matched_labels` (a dedicated tracking set) instead of `{spec.device.deviceInfo.label for spec in reconfig_specs}` as in your suggestion here. Both are logically equivalent, but `matched_labels` decouples the check from the mutated spec objects as you requested in the other comment.

libs/providers/vmware.py
krcmarik: Acknowledged — this is pre-existing tech debt (SHOULD rule, not MUST). Extracting helpers from a 338-line function carries significant regression risk without test execution capability and increases the scope of this PR. Will track as a separate refactoring task.

libs/providers/vmware.py
krcmarik: Acknowledged — `wait_task` raising `VmCloneError` for all vSphere task types is a pre-existing design issue (not introduced by this PR). Changing `wait_task`'s exception type or adding a parameter affects 10+ call sites across the VMWare provider. Out of scope for this MAC-preservation PR; will track separately.

libs/providers/vmware.py
krcmarik: Fixed: Added the `deviceInfo` ValueError to `clone_vm`'s Raises section.

**Note on approach:** Merged it into the existing `ValueError` entry (`"...or if a network device has no deviceInfo in the source VM config."`) rather than adding a second separate `ValueError:` line. This avoids duplicate exception type entries in the Raises section while documenting both cases.

libs/providers/vmware.py
krcmarik: Fixed: Introduced `matched_labels: set[str]` collected during the loop, used in the unmatched computation instead of deriving from mutated spec objects.

libs/providers/vmware.py
```diff
@@ -1336,21 +1338,28 @@ def clone_vm(
...
+                            raise ValueError(f"Device has no deviceInfo in VM '{source_vm_name}' config")
+                        nic_label = device.deviceInfo.label
+                        if device.addressType == "manual":
+                            # VMware generates lowercase MACs by default. Only uppercase manual MACs
+                            # need post-clone restoration — lowercase manual MACs get a new generated
+                            # MAC that already matches the expected case/format.
+                            if device.macAddress != device.macAddress.lower():
+                                uppercase_mac_nics.add(nic_label)
+                        device.addressType = "generated"
```
myakove, line 1356: **Mixed-case manual MACs get fully uppercased, not case-preserved**

The check `device.macAddress != device.macAddress.lower()` identifies any MAC containing uppercase characters (including mixed-case like `AA:bb:CC:dd:EE:ff`). However, the post-clone restoration at line 1534 does `dev.macAddress.upper()`, converting the **entire** MAC to uppercase.

If the original manual MAC was mixed-case, the restored MAC differs from the original — the "case preservation" promised in the PR title is actually "case uppercasing."

Suggestion: Store the original MAC alongside the label (e.g., `dict[str, str]` mapping `label → original_mac`) and restore the exact original MAC instead of blindly uppercasing:

```python
# Instead of set[str]
uppercase_mac_nics: dict[str, str] = {}  # label → original MAC

# When collecting:
if device.macAddress != device.macAddress.lower():
    uppercase_mac_nics[nic_label] = device.macAddress

# When restoring:
dev.macAddress = uppercase_mac_nics[dev.deviceInfo.label]  # exact original
```

libs/providers/vmware.py
```diff
@@ -1468,15 +1478,73 @@ def clone_vm(
             else:
                 raise
 
-        if res and self.fixture_store:
+        if self.fixture_store:
```
myakove, line 1481: **fixture_store registration misses retry failure path — potential VM leak**

If the retry `wait_task` call (line 1474) also fails and raises `VmCloneError`, the exception propagates out before reaching this fixture_store registration (line 1478). VMware may have partially created the VM during the retry, but it's never registered for teardown cleanup — causing a resource leak.

The pre-PR code had the same issue, but the refactoring here (unconditional registration) only covers the success path.

Suggestion: Register for teardown **before** the clone attempt so the name is always tracked regardless of success/failure:

```python
# Register clone name for teardown before attempting clone
if self.fixture_store:
    self.fixture_store["teardown"].setdefault(self.type, []).append({"name": clone_vm_name})

try:
    res = self.wait_task(...)
except VmCloneError as e:
    # retry logic...
```

Alternatively, add registration inside the retry `except` block before the second `wait_task`.

myakove (commented):
## Code Review

### :bulb: Suggestions (2)

| File | Line | Issue |
|------|------|-------|
| `libs/providers/vmware.py` | 1356 | **Mixed-case manual MACs get fully uppercased, not case-preserved** |
| `libs/providers/vmware.py` | 1481 | **fixture_store registration misses retry failure path — potential VM leak** |

---
*Review generated by pi*

libs/providers/vmware.py
krcmarik: VMware always stores manual MAC addresses in uppercase — mixed-case MACs (e.g., `AA:bb:CC:dd:EE:ff`) don't occur in practice. The `.upper()` restoration matches the actual format VMware uses. Storing and restoring the original MAC would add complexity for a scenario that doesn't exist with real vSphere VMs.

libs/providers/vmware.py
krcmarik: This was already addressed in a previous review round per your request — we removed the `res and` guard to register the clone unconditionally for teardown (`if self.fixture_store:` only). Moving the registration before the clone attempt would register VMs that may never have been created on vSphere, which could cause spurious cleanup failures. The current placement (after successful clone, before validation) balances leak prevention with correctness.

krcmarik:
@coderabbitai[bot]

The following review comments were reviewed and a decision was made:

### `libs/providers/vmware.py:1343` (outside diff) — HIGH: Fail fast when source hardware config is unavailable.
> Not addressed: Valid suggestion, but this guard (`if source_config and source_config.hardware and source_config.hardware.device:`) is pre-existing code not changed by this PR. The fail-fast pattern would be an improvement, but it's out of scope for this change.

redhat-qe-bot2:
**Clean rebase detected** — no code changes compared to previous head (`7f780b9`).
The following labels were preserved: `verified`, `commented-myakove`, `commented-krcmarik`, `commented-coderabbitai[bot]`, `lgtm-coderabbitai[bot]`.

libs/providers/vmware.py
```diff
@@ -1336,21 +1338,28 @@ def clone_vm(
             )
             device_changes.extend(disk_device_specs)
 
+        uppercase_mac_nics: set[str] = set()
         if regenerate_mac:
             source_config = source_vm.config
             if source_config and source_config.hardware and source_config.hardware.device:
                 for device in source_config.hardware.device:
                     if isinstance(device, vim.vm.device.VirtualEthernetCard):
+                        if not device.deviceInfo:
```
myakove, line 1347: 💡 **[SUGGESTION]** Duplicate `deviceInfo` validation pattern.

This same check-and-raise pattern appears in two places:
- Here (line 1347): `if not device.deviceInfo: raise ValueError(f"...VM '{source_vm_name}'...")`
- `_reconfigure_uppercase_manual_macs` (line 1525): `if not dev.deviceInfo: raise ValueError(f"...cloned VM '{clone_vm_name}'...")`

Per "No Duplicate Code (MUST)" — if the same logic appears twice, extract it.

Suggestion: Extract a small static helper:
```python
@staticmethod
def _get_nic_label(device: vim.vm.device.VirtualEthernetCard, vm_name: str) -> str:
    """Get NIC label from device, validating deviceInfo exists.

    Args:
        device (vim.vm.device.VirtualEthernetCard): The NIC device.
        vm_name (str): VM name for error context.

    Returns:
        str: The device label.

    Raises:
        ValueError: If device has no deviceInfo.
    """
    if not device.deviceInfo:
        raise ValueError(f"Device has no deviceInfo in VM '{vm_name}' config")
    return device.deviceInfo.label
```

Then use `nic_label = self._get_nic_label(device, source_vm_name)` at line 1349 and `label = self._get_nic_label(dev, clone_vm_name)` at line 1527.

---

myakove (commented):
## Code Review

Found **1** issue(s) in this PR:

### :bulb: Suggestions (1)

| File | Line | Issue |
|------|------|-------|
| `libs/providers/vmware.py` | 1347 | 💡 **[SUGGESTION]** Duplicate `deviceInfo` validation pattern. |

---
*Review generated by pi*

---

libs/providers/vmware.py
krcmarik: Skipped: Duplication is a 2-line pattern used twice, introduced by prior review feedback. Extracting a helper would expand 4 lines to ~14 — over-engineering per project guidelines.

libs/providers/vmware.py
```diff
@@ -1468,15 +1478,73 @@ def clone_vm(
...
+
+        unmatched = uppercase_mac_nics - matched_labels
+        if unmatched:
+            raise VmCloneError(f"NICs not found in cloned VM '{clone_vm_name}' for MAC restoration: {unmatched}")
+
+        if reconfig_specs:
+            self.wait_task(
+                task=vm.ReconfigVM_Task(spec=vim.vm.ConfigSpec(deviceChange=reconfig_specs)),
+                action_name=f"Reconfiguring manual MACs on {clone_vm_name}",
+            )
```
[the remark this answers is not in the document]
krcmarik: Not addressing — `power_on` defaults to `False` and is never called with `True` anywhere in the codebase. Cloned VMs are always created powered off, so `_reconfigure_uppercase_manual_macs()` always runs against a powered-off VM. The concern is valid in theory but doesn't apply in practice.

krcmarik:
@coderabbitai[bot]

The following review comments were reviewed and a decision was made:

### `utilities/post_migration.py:1683` (nitpick) — Broad `except Exception` in new nested-virt checks.
> Not addressed: Acknowledged — this is a false positive for this PR. PR #559 does not add or modify any code at lines 1683-1710 of `post_migration.py`; the only change is to `get_nic_by_mac` at line 867. The `except Exception` blocks referenced are pre-existing `check_vms()` aggregation patterns, not new code from this PR. Agree that a file-wide refactor is worth tracking separately.

rh-bot-1:

PR #563, opened 2026-06-18
MTV-5824: Add ca.crt provider secret field migration test

AmenB:
## Summary
- Add tier1 cold migration test verifying Forklift accepts the standard Kubernetes `ca.crt` field in provider secrets (MTV-4561 / kubev2v/forklift#5549)
- Add `ca_cert_key` parameter to `_fetch_and_store_cacert()` and `create_source_provider()` — defaults to `"cacert"`, no behavior change for existing tests
- Supported providers: vSphere, ESXi, RHV, OpenStack (skipped for OpenShift/OVA via collection-time hook)
- Uses existing `mtv-tests-rhel8` VM — no new infrastructure required

## Test plan
- [ ] Run `uv run pytest -s --collect-only -m tier1` and verify `TestCaCrtColdMigration` is collected
- [ ] Run with vSphere provider: confirm provider secret has `ca.crt` key (not `cacert`)
- [ ] Run regression alongside `TestSanityColdMtvMigration` to verify legacy `cacert` path is unaffected
- [ ] Verify test is skipped for OpenShift/OVA providers
- [ ] Verify test is skipped when `source_provider_insecure_skip_verify=true`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added support for storing provider CA certificates under a configurable secret field (e.g., `ca.crt`).
  * Added cold-migration planning coverage to validate CA certificate field readiness.

* **Bug Fixes**
  * Improved consistency of CA certificate handling across supported provider types.
  * Adjusted test collection to skip `ca_crt`-tagged tests for non-applicable provider kinds.

* **Tests**
  * Added new cold migration test verifying CA certificate secret contents plus readiness steps (storagemap → networkmap → plan).

* **Documentation**
  * Updated marker documentation and test guidance to include the new `ca_crt` category.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -784,6 +784,9 @@ class TestNameHere:
 - **Class-level parametrization**: Use `class_plan_config` with `indirect=True`
 - **Shared state**: Store resources on class with `self.__class__.attribute`
 - **Test ordering**: Use `@pytest.mark.incremental` at class level for sequential test dependencies
+- **4-step plan-readiness pattern**: verify_\<feature\> -> storagemap -> networkmap -> plan
+  Used when the feature under test is exercised during provider/plan creation (e.g., CA cert field
+  validation). No migration is executed — plan readiness proves the feature works.
 - **5-step pattern**: storagemap -> networkmap -> plan -> migrate -> check_vms
 - **6-step shared-disk pattern (Linux)**: storagemap -> networkmap -> plan -> migrate -> verify_shared_disk_data -> check_vms
   Shared disk tests insert `test_verify_shared_disk_data` before `test_check_vms`. This step mounts,
@@ -988,6 +991,7 @@ Class-scoped teardown fixture that cleans up migrated VMs after each test class 
 | `copyoffload`           | Copy-offload (XCOPY) tests             |
 | `copyoffload_sanity`    | Copy-offload sanity subset             |
 | `copyoffload_snapshots` | Copy-offload snapshot tests (vSphere)  |
+| `ca_crt`                | CA certificate field (ca.crt) tests    |
 | `vsphere`               | VMware vSphere provider-specific tests |
 | `shared_disk`           | Shared disk migration tests            |
 
@@ -997,6 +1001,7 @@ Provider-type skip logic runs in `pytest_collection_modifyitems` (in `conftest.p
 
 - Warm migration tests → `@pytest.mark.warm`
 - Copy-offload snapshot tests → `@pytest.mark.copyoffload_snapshots`
+- CA certificate field tests → `@pytest.mark.ca_crt`
 
 See the documented hook in `conftest.py` for how to add new provider-type skip rules.
 
--- a/conftest.py
+++ b/conftest.py
@@ -45,11 +45,7 @@ from utilities.copyoffload_constants import FORKLIFT_CONTROLLER_NAME
 from libs.base_provider import BaseProvider
 from libs.forklift_inventory import (
     ForkliftInventory,
-    OpenshiftForkliftInventory,
-    OpenstackForliftinventory,
-    OvaForkliftInventory,
-    OvirtForkliftInventory,
-    VsphereForkliftInventory,
+    create_forklift_inventory,
 )
 from libs.providers.openshift import OCPProvider
 from libs.providers.vmware import VMWareProvider
@@ -338,6 +334,19 @@ def pytest_collection_modifyitems(session, config, items):
                         or "vsphere" in item.keywords
                     ):
                         item.add_marker(vsphere_only_skip)
+
+            # Skip CA cert tests for providers that don't use CA certificates.
+            ca_cert_unsupported = (
+                Provider.ProviderType.OPENSHIFT,
+                Provider.ProviderType.OVA,
+            )
+            if source_provider_type in ca_cert_unsupported:
+                ca_cert_skip = pytest.mark.skip(
+                    reason=f"{source_provider_type} does not use CA certificates in provider secrets"
+                )
+                for item in items:
+                    if "ca_crt" in item.keywords:
+                        item.add_marker(ca_cert_skip)
 
     _session_store = get_fixture_store(session)
     vms_for_current_session: set = set()
@@ -1493,24 +1502,7 @@ def forklift_pods_state(ocp_admin_client: DynamicClient) -> None:
 def source_provider_inventory(
     ocp_admin_client: DynamicClient, mtv_namespace: str, source_provider: BaseProvider
 ) -> ForkliftInventory:
-    if not source_provider.ocp_resource:
-        raise ValueError("source_provider.ocp_resource is not set")
-
-    providers = {
-        Provider.ProviderType.OVA: OvaForkliftInventory,
-        Provider.ProviderType.RHV: OvirtForkliftInventory,
-        Provider.ProviderType.VSPHERE: VsphereForkliftInventory,
-        Provider.ProviderType.OPENSHIFT: OpenshiftForkliftInventory,
-        Provider.ProviderType.OPENSTACK: OpenstackForliftinventory,
-    }
-    provider_instance = providers.get(source_provider.type)
-
-    if not provider_instance:
-        raise ValueError(f"Provider {source_provider.type} not implemented")
-
-    return provider_instance(  # type: ignore
-        client=ocp_admin_client, namespace=mtv_namespace, provider_name=source_provider.ocp_resource.name
-    )
+    return create_forklift_inventory(client=ocp_admin_client, mtv_namespace=mtv_namespace, provider=source_provider)
 
 
 @pytest.fixture(scope="session")
--- a/libs/forklift_inventory.py
+++ b/libs/forklift_inventory.py
@@ -1,5 +1,7 @@
+from __future__ import annotations
+
 import abc
-from typing import Any
+from typing import TYPE_CHECKING, Any
 
 from kubernetes.dynamic.client import DynamicClient
 from ocp_resources.persistent_volume_claim import PersistentVolumeClaim
@@ -8,7 +10,59 @@ from ocp_resources.route import Route
 from simple_logger.logger import get_logger
 from timeout_sampler import TimeoutExpiredError, TimeoutSampler
 
+if TYPE_CHECKING:
+    from libs.base_provider import BaseProvider
+
+PROVIDER_INVENTORY_MAP: dict[str, type[ForkliftInventory]] = {}
+
 LOGGER = get_logger(__name__)
+
+
+def _register_inventory_classes() -> None:
+    """Populate PROVIDER_INVENTORY_MAP after all classes are defined."""
+    PROVIDER_INVENTORY_MAP.update({
+        Provider.ProviderType.OVA: OvaForkliftInventory,
+        Provider.ProviderType.RHV: OvirtForkliftInventory,
+        Provider.ProviderType.VSPHERE: VsphereForkliftInventory,
+        Provider.ProviderType.OPENSHIFT: OpenshiftForkliftInventory,
+        Provider.ProviderType.OPENSTACK: OpenstackForliftinventory,
+    })
+
+
+def create_forklift_inventory(
+    client: DynamicClient,
+    mtv_namespace: str,
+    provider: BaseProvider,
+) -> ForkliftInventory:
+    """ForkliftInventory instance for the given provider.
+
+    Args:
+        client (DynamicClient): OpenShift admin client.
+        mtv_namespace (str): MTV operator namespace.
+        provider (BaseProvider): Source provider instance.
+
+    Returns:
+        ForkliftInventory: Inventory instance matching the provider type.
+
+    Raises:
+        ValueError: If ocp_resource is not set on the provider.
+        ValueError: If the provider type is not supported.
+    """
+    if not PROVIDER_INVENTORY_MAP:
+        _register_inventory_classes()
+
+    if provider.ocp_resource is None:
+        raise ValueError(f"{provider.type} provider ocp_resource is not set")
+
+    provider_class = PROVIDER_INVENTORY_MAP.get(provider.type)
+    if provider_class is None:
+        raise ValueError(f"Provider {provider.type} not implemented")
+
+    return provider_class(  # type: ignore[call-arg]  # Subclasses use 'namespace' param while base class uses 'mtv_namespace'
+        client=client,
+        namespace=mtv_namespace,
+        provider_name=provider.ocp_resource.name,
+    )
 
 
 class ForkliftInventory(abc.ABC):
--- a/pytest.ini
+++ b/pytest.ini
@@ -31,6 +31,7 @@ markers =
     openstack: OpenStack provider-specific tests
     openshift: OpenShift provider-specific tests
     deep_inspection: Deep Inspection / Conversion CR tests (vSphere only)
+    ca_crt: CA certificate field (ca.crt) provider secret tests
     upgrade: MTV operator upgrade tests
 
 junit_logging = all
--- a/tests/cold/conftest.py
+++ b/tests/cold/conftest.py
@@ -1,0 +1,81 @@
+from __future__ import annotations
+
+from collections.abc import Generator
+from typing import TYPE_CHECKING, Any
+
+import pytest
+
+from libs.base_provider import BaseProvider
+from libs.forklift_inventory import ForkliftInventory, create_forklift_inventory
+from utilities.utils import create_source_provider, get_value_from_py_config
+
+if TYPE_CHECKING:
+    from kubernetes.dynamic import DynamicClient
+    from ocp_resources.secret import Secret
+
+
+@pytest.fixture(scope="class")  # Class-scoped: separate provider per test class with ca.crt in the secret
+def ca_crt_source_provider(
+    fixture_store: dict[str, Any],
+    session_uuid: str,
+    source_provider_data: dict[str, Any],
+    target_namespace: str,
+    ocp_admin_client: DynamicClient,
+    tmp_path_factory: pytest.TempPathFactory,
+    destination_ocp_secret: Secret,  # pragma: allowlist secret
+) -> Generator[BaseProvider, None, None]:
+    """Source provider configured with ca.crt secret field instead of cacert.
+
+    Uses the standard Kubernetes ca.crt convention (MTV-4561) to verify
+    Forklift accepts the new field name for CA certificates.
+
+    Args:
+        fixture_store (dict[str, Any]): Session fixture store for resource tracking.
+        session_uuid (str): Unique session identifier.
+        source_provider_data (dict[str, Any]): Provider configuration from providers JSON.
+        target_namespace (str): Target namespace for provider resources.
+        ocp_admin_client (DynamicClient): OpenShift admin client.
+        tmp_path_factory (pytest.TempPathFactory): Temp directory factory for cert files.
+        destination_ocp_secret (Secret): Destination OCP cluster secret.
+
+    Yields:
+        BaseProvider: Source provider instance using ca.crt in the secret.
+    """
+    with create_source_provider(
+        fixture_store=fixture_store,
+        session_uuid=session_uuid,
+        source_provider_data=source_provider_data,
+        namespace=target_namespace,
+        admin_client=ocp_admin_client,
+        tmp_dir=tmp_path_factory,
+        ocp_admin_client=ocp_admin_client,
+        destination_ocp_secret=destination_ocp_secret,
+        insecure=get_value_from_py_config(value="source_provider_insecure_skip_verify"),
+        ca_cert_key="ca.crt",
+    ) as _source_provider:
+        yield _source_provider
+
+    _source_provider.disconnect()
+
+
+@pytest.fixture(scope="class")  # Class-scoped: must match ca_crt_source_provider scope
+def ca_crt_source_provider_inventory(
+    ocp_admin_client: DynamicClient,
+    mtv_namespace: str,
+    ca_crt_source_provider: BaseProvider,
+) -> ForkliftInventory:
+    """ForkliftInventory instance for the ca.crt provider.
+
+    Args:
+        ocp_admin_client (DynamicClient): OpenShift admin client.
+        mtv_namespace (str): MTV operator namespace.
+        ca_crt_source_provider (BaseProvider): Source provider using ca.crt secret field.
+
+    Returns:
+        ForkliftInventory: Inventory instance for the ca.crt provider.
+    """
+    return create_forklift_inventory(
+        client=ocp_admin_client,
+        mtv_namespace=mtv_namespace,
+        provider=ca_crt_source_provider,
+    )
--- a/tests/cold/test_ca_crt_migration.py
+++ b/tests/cold/test_ca_crt_migration.py
@@ -1,0 +1,153 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Any
+
+import pytest
+from ocp_resources.network_map import NetworkMap
+from ocp_resources.plan import Plan
+from ocp_resources.secret import Secret
+from ocp_resources.storage_map import StorageMap
+from pytest_testconfig import config as py_config
+
+from libs.base_provider import BaseProvider
+from libs.forklift_inventory import ForkliftInventory
+from utilities.mtv_migration import (
+    create_plan_resource,
+    get_network_migration_map,
+    get_storage_migration_map,
+)
+from utilities.utils import get_value_from_py_config, populate_vm_ids
+
+if TYPE_CHECKING:
+    from kubernetes.dynamic import DynamicClient
+
+    from libs.providers.openshift import OCPProvider
+
+
+@pytest.mark.vsphere
+@pytest.mark.rhv
+@pytest.mark.openstack
+@pytest.mark.esxi
+@pytest.mark.ca_crt
+@pytest.mark.tier1
+@pytest.mark.incremental
+@pytest.mark.skipif(
+    get_value_from_py_config("source_provider_insecure_skip_verify"),
+    reason="CA cert field test requires SSL verification enabled (insecure=false)",
+)
+@pytest.mark.parametrize(
+    "class_plan_config",
+    [
+        pytest.param(
+            py_config["tests_params"]["test_ca_crt_cold_migration"],
+        )
+    ],
+    indirect=True,
+    ids=["MTV-4561-ca-crt"],
+)
+class TestCaCrtColdMigration:
+    """Verify Forklift accepts ca.crt secret field instead of cacert (MTV-4561).
+
+    Validates provider connectivity and plan readiness using the standard
+    Kubernetes ca.crt convention. No migration is executed — the CA cert
+    code path (getTLSConfig -> Authenticate) runs during provider creation,
+    so plan readiness proves the feature works.
+    """
+
+    storage_map: StorageMap
+    network_map: NetworkMap
+    plan_resource: Plan
+
+    @pytest.mark.usefixtures("prepared_plan")
+    def test_verify_ca_crt_secret(
+        self,
+        ca_crt_source_provider: BaseProvider,
+        ocp_admin_client: DynamicClient,
+    ) -> None:
+        """Verify the provider secret uses ca.crt field instead of cacert."""
+        assert ca_crt_source_provider.ocp_resource is not None, "ocp_resource is not set"
+        secret_ref = ca_crt_source_provider.ocp_resource.instance.spec.secret
+        secret = Secret(
+            client=ocp_admin_client,
+            name=secret_ref.name,
+            namespace=secret_ref.namespace,
+        )
+        secret_data_keys = list(secret.instance.data.keys())
+        assert "ca.crt" in secret_data_keys, f"Expected 'ca.crt' in secret keys, got: {secret_data_keys}"
+        assert "cacert" not in secret_data_keys, (
+            f"Legacy 'cacert' should not be in secret keys, got: {secret_data_keys}"
+        )
+
+    def test_create_storagemap(
+        self,
+        prepared_plan: dict[str, Any],
+        fixture_store: dict[str, Any],
+        ocp_admin_client: DynamicClient,
+        ca_crt_source_provider: BaseProvider,
+        destination_provider: OCPProvider,
+        ca_crt_source_provider_inventory: ForkliftInventory,
+        target_namespace: str,
+    ) -> None:
+        """Create StorageMap resource for migration."""
+        vms = [vm["name"] for vm in prepared_plan["virtual_machines"]]
+        self.__class__.storage_map = get_storage_migration_map(
+            fixture_store=fixture_store,
+            source_provider=ca_crt_source_provider,
+            destination_provider=destination_provider,
+            source_provider_inventory=ca_crt_source_provider_inventory,
+            ocp_admin_client=ocp_admin_client,
+            target_namespace=target_namespace,
+            vms=vms,
+        )
+        assert self.storage_map, "StorageMap creation failed"
+
+    def test_create_networkmap(
+        self,
+        prepared_plan: dict[str, Any],
+        fixture_store: dict[str, Any],
+        ocp_admin_client: DynamicClient,
+        ca_crt_source_provider: BaseProvider,
+        destination_provider: OCPProvider,
+        ca_crt_source_provider_inventory: ForkliftInventory,
+        target_namespace: str,
+        multus_network_name: dict[str, str],
+    ) -> None:
+        """Create NetworkMap resource for migration."""
+        vms = [vm["name"] for vm in prepared_plan["virtual_machines"]]
+        self.__class__.network_map = get_network_migration_map(
+            fixture_store=fixture_store,
+            source_provider=ca_crt_source_provider,
+            destination_provider=destination_provider,
+            source_provider_inventory=ca_crt_source_provider_inventory,
+            ocp_admin_client=ocp_admin_client,
+            target_namespace=target_namespace,
+            multus_network_name=multus_network_name,
+            vms=vms,
+        )
+        assert self.network_map, "NetworkMap creation failed"
+
+    def test_create_plan(
+        self,
+        prepared_plan: dict[str, Any],
+        fixture_store: dict[str, Any],
+        ocp_admin_client: DynamicClient,
+        ca_crt_source_provider: BaseProvider,
+        destination_provider: OCPProvider,
+        target_namespace: str,
+        ca_crt_source_provider_inventory: ForkliftInventory,
+    ) -> None:
+        """Create MTV Plan CR resource."""
+        populate_vm_ids(prepared_plan, ca_crt_source_provider_inventory)
+
+        self.__class__.plan_resource = create_plan_resource(
+            ocp_admin_client=ocp_admin_client,
+            fixture_store=fixture_store,
+            source_provider=ca_crt_source_provider,
+            destination_provider=destination_provider,
+            storage_map=self.storage_map,
+            network_map=self.network_map,
+            virtual_machines_list=prepared_plan["virtual_machines"],
+            target_namespace=target_namespace,
+            warm_migration=prepared_plan.get("warm_migration", False),
+        )
+        assert self.plan_resource, "Plan creation failed"
--- a/utilities/utils.py
+++ b/utilities/utils.py
@@ -160,21 +160,23 @@ def _fetch_and_store_cacert(
     secret_string_data: dict[str, Any],
     tmp_dir: pytest.TempPathFactory | None,
     session_uuid: str,
+    ca_cert_key: str = "cacert",
 ) -> Path:
-    """
-    Fetch CA certificate from provider and store in secret data.
+    """Fetch CA certificate from provider and store in secret data.
 
     Args:
-        source_provider_data: Provider configuration with 'type' and 'fqdn'
-        secret_string_data: Secret data dict to add 'cacert' to
-        tmp_dir: Temp directory factory for cert file
-        session_uuid: Session UUID for unique filename
+        source_provider_data (dict[str, Any]): Provider configuration with 'type' and 'fqdn'.
+        secret_string_data (dict[str, Any]): Secret data dict to store the certificate in.
+        tmp_dir (pytest.TempPathFactory | None): Temp directory factory for cert file.
+        session_uuid (str): Session UUID for unique filename.
+        ca_cert_key (str): Secret field name for the CA certificate. Defaults to "cacert".
+            Use "ca.crt" for the standard Kubernetes convention (MTV-4561).
 
     Returns:
-        Path to the certificate file
+        Path: Path to the certificate file.
 
     Raises:
-        ValueError: If tmp_dir is not provided
+        ValueError: If tmp_dir is not provided.
     """
     source_provider_type = source_provider_data["type"]
     if not tmp_dir:
@@ -184,7 +186,7 @@ def _fetch_and_store_cacert(
         provider_fqdn=source_provider_data["fqdn"],
         cert_file=tmp_dir.mktemp(source_provider_type.upper()) / f"{source_provider_type}_{session_uuid}_cert.crt",
     )
-    secret_string_data["cacert"] = cert_file.read_text()
+    secret_string_data[ca_cert_key] = cert_file.read_text()
     return cert_file
 
 
@@ -273,7 +275,30 @@ def create_source_provider(
     destination_ocp_secret: Secret,
     insecure: bool,
     tmp_dir: pytest.TempPathFactory | None = None,
+    ca_cert_key: str = "cacert",
 ) -> Generator[BaseProvider, None, None]:
+    """Source provider configured and registered as a Forklift Provider CR.
+
+    Args:
+        source_provider_data (dict[str, Any]): Provider configuration from providers JSON.
+        namespace (str): Target namespace for the provider resources.
+        admin_client (DynamicClient): OpenShift client for resource creation.
+        session_uuid (str): Unique session identifier for resource naming.
+        fixture_store (dict[str, Any]): Session fixture store for resource tracking and teardown.
+        ocp_admin_client (DynamicClient): OpenShift admin client for cluster operations.
+        destination_ocp_secret (Secret): Secret for the destination OCP provider.
+        insecure (bool): Whether to skip TLS certificate verification.
+        tmp_dir (pytest.TempPathFactory | None): Temporary directory factory for storing fetched certificates.
+        ca_cert_key (str): Secret data key for the CA certificate (default: "cacert",
+            use "ca.crt" to test the alternative secret field name).
+
+    Yields:
+        BaseProvider: Connected source provider instance.
+
+    Raises:
+        ValueError: If the provider type cannot be determined from source_provider_data.
+        ValueError: If the provider secret fails to create.
+    """
     # common
     source_provider_secret: Secret | None = None
     source_provider: Any = None
@@ -320,7 +345,7 @@ def create_source_provider(
                 LOGGER.info(f"Normalized VMware API URL for TLS: '{original_url}' -> '{normalized_url}'")
             source_provider_data_copy["api_url"] = normalized_url
             secret_string_data["url"] = normalized_url
-            _fetch_and_store_cacert(source_provider_data_copy, secret_string_data, tmp_dir, session_uuid)
+            _fetch_and_store_cacert(source_provider_data_copy, secret_string_data, tmp_dir, session_uuid, ca_cert_key)
 
     elif rhv_provider(provider_data=source_provider_data_copy):
         source_provider = OvirtProvider
@@ -330,7 +355,9 @@ def create_source_provider(
 
         # Always fetch CA certificate for RHV provider, even when insecure=True
         # The certificate is required for imageio connection, insecureSkipVerify controls validation
-        cert_file = _fetch_and_store_cacert(source_provider_data_copy, secret_string_data, tmp_dir, session_uuid)
+        cert_file = _fetch_and_store_cacert(
+            source_provider_data_copy, secret_string_data, tmp_dir, session_uuid, ca_cert_key
+        )
 
         # Set ca_file in provider_args only when secure mode (for SDK connection)
         if not insecure:
@@ -355,7 +382,7 @@ def create_source_provider(
 
         # Add CA certificate for SSL verification
         if not insecure:
-            _fetch_and_store_cacert(source_provider_data_copy, secret_string_data, tmp_dir, session_uuid)
+            _fetch_and_store_cacert(source_provider_data_copy, secret_string_data, tmp_dir, session_uuid, ca_cert_key)
 
     elif ova_provider(provider_data=source_provider_data_copy):
         source_provider = OVAProvider
--- a/README.md
+++ b/README.md
@@ -272,6 +272,7 @@ The Quick Start runs **tier0** tests (smoke tests). You can run other test categ
 | `copyoffload_sanity` | Copy-offload sanity subset (see below) | Quick copy-offload validation |
 | `warm` | Warm migrations (VMs stay running) | Specific scenario testing |
 | `shared_disk` | Shared disk migration tests | Testing shared disk between VMs |
+| `ca_crt` | CA certificate field (ca.crt) in provider secrets | Testing ca.crt secret field support |
 | `vsphere` | VMware vSphere provider-specific tests | Tests specific to vSphere provider |
 | `upgrade` | Migration across MTV operator upgrades | Validating upgrade compatibility |
 
--- a/tests/tests_config/config.py
+++ b/tests/tests_config/config.py
@@ -721,6 +721,12 @@ tests_params: dict = {
         "migrate_shared_disks": True,
         "target_power_state": "on",
     },
+    "test_ca_crt_cold_migration": {
+        "virtual_machines": [
+            {"name": "mtv-tests-rhel8", "guest_agent": True},
+        ],
+        "warm_migration": False,
+    },
     "test_upgrade_cold_migration": {
         "virtual_machines": [
             {"name": "mtv-tests-rhel8", "guest_agent": True},


tests/cold/conftest.py
```diff
@@ -0,0 +1,79 @@
...
+    target_namespace: str,
+    ocp_admin_client: "DynamicClient",
+    tmp_path_factory: pytest.TempPathFactory,
+    destination_ocp_secret: "Secret",
+) -> Generator[BaseProvider, None, None]:
+    """Create a source provider with ca.crt secret field instead of cacert.
+
+    Uses the standard Kubernetes ca.crt convention (MTV-4561) to verify
+    Forklift accepts the new field name for CA certificates.
+    """
```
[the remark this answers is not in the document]
AmenB: Fixed in b04ff09 — added full Args/Yields/Returns/Raises sections to both fixture docstrings.

For the test methods: they follow the existing project pattern (e.g. `test_mtv_cold_migration.py`) where test methods have single-line docstrings. Per CLAUDE.md, Args/Returns/Raises are included only "as applicable" — test methods don't return values or raise exceptions (they assert).

tests/cold/conftest.py
```diff
@@ -0,0 +1,79 @@
...
+
+    providers: dict[str, type[ForkliftInventory]] = {
+        Provider.ProviderType.RHV: OvirtForkliftInventory,
+        Provider.ProviderType.VSPHERE: VsphereForkliftInventory,
+        Provider.ProviderType.OPENSTACK: OpenstackForliftinventory,
+    }
+    provider_class = providers.get(ca_crt_source_provider.type)
+
+    if not provider_class:
+        raise ValueError(f"Provider {ca_crt_source_provider.type} does not use CA certificates")
```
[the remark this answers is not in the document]
AmenB: Fixed in b04ff09 — changed `if not ca_crt_source_provider.ocp_resource:` to `if ca_crt_source_provider.ocp_resource is None:` for explicit None check per CLAUDE.md.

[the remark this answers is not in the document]
AmenB: Fixed in 7498a02 — both checks now use explicit `is None`. The first (`ocp_resource`) was fixed in b04ff09, the second (`provider_class`) in 7498a02.

tests/cold/test_ca_crt_migration.py
```diff
@@ -0,0 +1,156 @@
...
+            plan=prepared_plan,
+            source_provider=ca_crt_source_provider,
+            destination_provider=destination_provider,
+            network_map_resource=self.network_map,
+            storage_map_resource=self.storage_map,
+            source_provider_data=source_provider_data,
+            source_vms_namespace=source_vms_namespace,
+            source_provider_inventory=ca_crt_source_provider_inventory,
+            vm_ssh_connections=vm_ssh_connections,
+        )
```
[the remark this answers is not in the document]
AmenB: Fixed in 8f44f23 — added complete type annotations (parameter types + `-> None`) to all 5 test methods.

tests/cold/test_ca_crt_migration.py
```diff
@@ -0,0 +1,156 @@
...
+        self,
+        prepared_plan,
+        ca_crt_source_provider,
+        destination_provider,
+        source_provider_data,
+        target_namespace,
+        source_vms_namespace,
+        ca_crt_source_provider_inventory,
+        vm_ssh_connections,
+    ):
```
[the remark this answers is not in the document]
AmenB: Fixed in 8f44f23 — removed unused `target_namespace` from `test_check_vms`.

tests/cold/conftest.py
```diff
@@ -0,0 +1,103 @@
...
+
+    providers = {
+        Provider.ProviderType.RHV: OvirtForkliftInventory,
+        Provider.ProviderType.VSPHERE: VsphereForkliftInventory,
+        Provider.ProviderType.OPENSTACK: OpenstackForliftinventory,
+    }
+    provider_class = providers.get(ca_crt_source_provider.type)
+
+    if provider_class is None:
+        raise ValueError(f"Provider {ca_crt_source_provider.type} does not use CA certificates")
```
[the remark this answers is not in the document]
AmenB: Not a real issue — ESXi providers use `Provider.ProviderType.VSPHERE` as their type. There is no separate `ESXI` provider type. The `@pytest.mark.esxi` marker is for test filtering only, not a distinct provider type.

The session-scoped `source_provider_inventory` fixture (conftest.py:1503) also has no ESXI entry for the same reason.

tests/cold/conftest.py
```diff
@@ -0,0 +1,105 @@
...
+
+    Raises:
+        ValueError: If ocp_resource is not set or provider type doesn't use CA certificates.
+    """
+    if ca_crt_source_provider.ocp_resource is None:
+        raise ValueError("ca_crt_source_provider.ocp_resource is not set")
+
+    providers: dict[
+        str, type[OvirtForkliftInventory] | type[VsphereForkliftInventory] | type[OpenstackForliftinventory]
+    ] = {
```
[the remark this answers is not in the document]
AmenB: The union type is intentional — `type[ForkliftInventory]` was the original annotation but mypy rejects it because `ForkliftInventory` is abstract (`Cannot instantiate abstract class`). The union of concrete subclasses is the fix for that mypy error. This is a typing constraint, not a style choice.

tests/cold/test_ca_crt_migration.py
```diff
@@ -0,0 +1,166 @@
...
+            ocp_admin_client=ocp_admin_client,
+            fixture_store=fixture_store,
+            source_provider=ca_crt_source_provider,
+            destination_provider=destination_provider,
+            storage_map=self.storage_map,
+            network_map=self.network_map,
+            virtual_machines_list=prepared_plan["virtual_machines"],
+            target_namespace=target_namespace,
+            warm_migration=prepared_plan.get("warm_migration", False),
+        )
```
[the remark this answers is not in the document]
AmenB: Not applicable — CLAUDE.md explicitly exempts `warm_migration` as an optional flag where `.get()` is correct:

> `Correct — .get() for optional feature flags (not present in every test config)`
> `warm_migration = plan.get("warm_migration", False)`
> `copyoffload = plan.get("copyoffload", False)`

This follows the same pattern used in all other test files (e.g. `test_mtv_cold_migration.py:112`).

tests/cold/conftest.py
```diff
@@ -0,0 +1,105 @@
...
+        insecure=get_value_from_py_config(value="source_provider_insecure_skip_verify"),
+        ca_cert_key="ca.crt",
+    ) as _source_provider:
+        yield _source_provider
+
+    _source_provider.disconnect()
+
+
+@pytest.fixture(scope="class")
+def ca_crt_source_provider_inventory(
```
myakove, line 62: **[WARNING]** `ca_crt_source_provider_inventory` duplicates session `source_provider_inventory` logic.

This fixture is nearly identical to the root `conftest.py` `source_provider_inventory` (lines 1484-1505). The only differences: (1) takes `ca_crt_source_provider` instead of `source_provider`, (2) excludes OVA/OpenShift from the providers dict. The core logic (lookup provider class, validate, instantiate) is copy-pasted.

**Suggestion:** Extract a shared helper (e.g., `_create_forklift_inventory(client, namespace, provider)`) in `utilities/` that both fixtures call, parameterized by the provider instance. The CA-cert fixture can filter out unsupported types before calling.

tests/cold/conftest.py
```diff
@@ -0,0 +1,105 @@
...
+    target_namespace: str,
+    ocp_admin_client: "DynamicClient",
+    tmp_path_factory: pytest.TempPathFactory,
+    destination_ocp_secret: "Secret",  # pragma: allowlist secret
+) -> Generator[BaseProvider, None, None]:
+    """Create a source provider with ca.crt secret field instead of cacert.
+
+    Uses the standard Kubernetes ca.crt convention (MTV-4561) to verify
+    Forklift accepts the new field name for CA certificates.
+
```
myakove, line 31: **[WARNING]** Fixture docstrings use verb "Create" instead of noun form.

Per AGENTS.md fixture naming convention, docstrings should describe what the fixture IS, not what it DOES.
- Line 37: `"Create a source provider with..."` → `"Source provider configured with ca.crt secret field instead of cacert."`
- Line 73: `"Create a ForkliftInventory for..."` → `"ForkliftInventory instance for the ca.crt provider."`

tests/cold/conftest.py
```diff
@@ -0,0 +1,105 @@
...
+    VsphereForkliftInventory,
+)
+from utilities.utils import create_source_provider, get_value_from_py_config
+
+if TYPE_CHECKING:
+    from kubernetes.dynamic import DynamicClient
+    from ocp_resources.secret import Secret
+
+
+@pytest.fixture(scope="class")
```
myakove: **[WARNING]** Class-scoped fixture lacks comment explaining scope choice.

The session-scoped `source_provider` and `source_provider_inventory` fixtures it mirrors are session-scoped. Per AGENTS.md "Fixture Scope Changes (MUST)", scope differences from existing patterns should have a comment explaining why.

**Suggestion:** Add a comment, e.g.:
```python
@pytest.fixture(scope="class")  # Class-scoped: creates a separate provider per test class with ca.crt key
```

tests/cold/conftest.py
```diff
@@ -0,0 +1,105 @@
...
+    providers: dict[
+        str, type[OvirtForkliftInventory] | type[VsphereForkliftInventory] | type[OpenstackForliftinventory]
+    ] = {
+        Provider.ProviderType.RHV: OvirtForkliftInventory,
+        Provider.ProviderType.VSPHERE: VsphereForkliftInventory,
+        Provider.ProviderType.OPENSTACK: OpenstackForliftinventory,
+    }
+    provider_class = providers.get(ca_crt_source_provider.type)
+
+    if provider_class is None:
```
myakove: **[WARNING]** `ca_crt_source_provider_inventory` raises `ValueError` for unsupported provider types instead of skipping.

If run against an OVA or OpenShift provider, the fixture raises `ValueError`. The test has `@pytest.mark.vsphere`, `@pytest.mark.rhv`, `@pytest.mark.openstack` markers but no collection-time skip logic for OVA/OpenShift.

**Suggestion:** Use `pytest.skip(f"Provider {ca_crt_source_provider.type} does not use CA certificates")` instead of `raise ValueError`, or add `@pytest.mark.skipif` at class level to prevent collection on unsupported providers.

tests/cold/conftest.py
```diff
@@ -0,0 +1,105 @@
...
+        mtv_namespace (str): MTV operator namespace.
+        ca_crt_source_provider (BaseProvider): Source provider using ca.crt secret field.
+
+    Returns:
+        ForkliftInventory: Inventory instance for the ca.crt provider.
+
+    Raises:
+        ValueError: If ocp_resource is not set or provider type doesn't use CA certificates.
+    """
+    if ca_crt_source_provider.ocp_resource is None:
```
myakove: **[SUGGESTION]** Raises section in docstring lists `ValueError` for two different conditions without distinguishing.

Currently: `"Raises: ValueError: If ocp_resource is not set or provider type doesn't use CA certificates."`

**Suggestion:** Split into two entries:
```python
Raises:
    ValueError: If ocp_resource is not set on the provider.
    ValueError: If the provider type does not support CA certificates.
```

myakove (commented):
## Code Review

### :bulb: Suggestions (5)

| File | Line | Issue |
|------|------|-------|
| `tests/cold/conftest.py` | 68 | **[WARNING]** `ca_crt_source_provider_inventory` duplicates session `source_prov |
| `tests/cold/conftest.py` | 37 | **[WARNING]** Fixture docstrings use verb "Create" instead of noun form. |
| `tests/cold/conftest.py` | 23 | **[WARNING]** Class-scoped fixture lacks comment explaining scope choice. |
| `tests/cold/conftest.py` | 98 | **[WARNING]** `ca_crt_source_provider_inventory` raises `ValueError` for unsuppo |
| `tests/cold/conftest.py` | 86 | **[SUGGESTION]** Raises section in docstring lists `ValueError` for two differen |

---
*Review generated by pi*

tests/cold/test_ca_crt_migration.py
```diff
@@ -0,0 +1,166 @@
...
+    ],
+    indirect=True,
+    ids=["MTV-4561-ca-crt"],
+)
+@pytest.mark.usefixtures("cleanup_migrated_vms")
+class TestCaCrtColdMigration:
+    """Cold migration using ca.crt secret field instead of cacert (MTV-4561)."""
+
+    storage_map: StorageMap
```
krcmarik, line 57: We discussed similar test offline some time ago (the skipping tls verification flag test) — the test runs a full 5-step migration just to verify that Forklift accepts `ca.crt` as the secret field name. That's overkill. Once the Plan CR reaches Ready, you've already proven:

1. The secret with `ca.crt` was accepted
2. The provider connected and inventory populated (Plan can't resolve VMs without it)

The actual migration and post-migration checks don't add anything to the `ca.crt` validation — that's already covered by `TestSanityColdMtvMigration`.

I'd stop at `test_create_plan` (3-step pattern) and drop `cleanup_migrated_vms`, `vm_ssh_connections`, etc. Much faster, same coverage.

AmenB: Hey @krcmarik 
We discussed this and agree on saving resources for tier0 — but this is a tier1 test where execution
time isn't a constraint.

Also, while running the full 5-step migration during development, we discovered MTV-5867 — the OpenStack
client was not using GetCACert() from PR #5549, so ca.crt was silently ignored and Forklift fell back
to the system CA pool. The provider reached Ready and the Plan reached Ready, but the controller logs
showed "CA certificate was not provided, system CA cert pool is used". We only caught this by running
the full migration and inspecting the controller logs — a 3-step test would have passed silently with
the same bug present.

The fix PR: https://github.com/kubev2v/forklift/pull/7198

That said, neither 3-step nor 5-step can programmatically detect this today. But if we add log analysis
for passed scenarios in the future (e.g., scanning controller logs for unexpected fallback messages),
the full migration provides more surface area to catch corner cases like this. Better to have the
coverage in place now.

What do you think, should we keep it this way or cut it to 3 steps as you suggested?

tests/cold/conftest.py
AmenB: Fixed — extracted `create_forklift_inventory()` shared helper in `libs/forklift_inventory.py`. Both the session `source_provider_inventory` and class-scoped `ca_crt_source_provider_inventory` now call it.

tests/cold/conftest.py
AmenB: Fixed — changed to noun form:
- `"Source provider configured with ca.crt secret field instead of cacert."`
- `"ForkliftInventory instance for the ca.crt provider."`

tests/cold/conftest.py
AmenB: Fixed — added inline comments:
```python
@pytest.fixture(scope="class")  # Class-scoped: separate provider per test class with ca.crt in the secret
@pytest.fixture(scope="class")  # Class-scoped: must match ca_crt_source_provider scope
```

tests/cold/conftest.py
AmenB: Fixed — the fixture now delegates to the shared `create_forklift_inventory()` helper which includes all provider types (OVA, OpenShift included). No more provider-type filtering or ValueError for unsupported types in the fixture itself.

tests/cold/conftest.py
AmenB: Fixed — simplified to one entry since validation moved to the shared helper:
```python
Raises:
    ValueError: If ocp_resource is not set on the provider.
```

libs/forklift_inventory.py
```diff
@@ -8,8 +10,60 @@
 from simple_logger.logger import get_logger
 from timeout_sampler import TimeoutExpiredError, TimeoutSampler
 
+if TYPE_CHECKING:
+    from libs.base_provider import BaseProvider
+
 LOGGER = get_logger(__name__)
 
+PROVIDER_INVENTORY_MAP: dict[str, type["ForkliftInventory"]] = {}
```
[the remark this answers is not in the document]
AmenB: Fixed — moved `PROVIDER_INVENTORY_MAP` before `LOGGER`, right after imports.

tests/cold/conftest.py
```diff
@@ -0,0 +1,84 @@
...
+@pytest.fixture(scope="class")  # Class-scoped: separate provider per test class with ca.crt in the secret
+def ca_crt_source_provider(
+    fixture_store: dict[str, Any],
+    session_uuid: str,
+    source_provider_data: dict[str, Any],
+    target_namespace: str,
+    ocp_admin_client: "DynamicClient",
+    tmp_path_factory: pytest.TempPathFactory,
+    destination_ocp_secret: "Secret",  # pragma: allowlist secret
+) -> Generator[BaseProvider, None, None]:
```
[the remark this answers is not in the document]
AmenB: Not applicable — `dict[str, Any]` is the standard type for `fixture_store`, `source_provider_data`, and `prepared_plan` throughout the entire codebase. These are genuinely dynamic dicts with mixed-type values. Every existing fixture and test method uses the same pattern.

tests/cold/test_ca_crt_migration.py
```diff
@@ -0,0 +1,166 @@
...
+@pytest.mark.parametrize(
+    "class_plan_config",
+    [
+        pytest.param(
+            py_config["tests_params"]["test_ca_crt_cold_migration"],
+        )
+    ],
+    indirect=True,
+    ids=["MTV-4561-ca-crt"],
+)
```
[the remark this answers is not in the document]
AmenB: The shared `create_forklift_inventory()` helper now includes all provider types (OVA, OpenShift included), so the fixture won't crash for any provider. For OVA/OpenShift the test will pass but test nothing meaningful since those providers don't use CA certs — this is acceptable for a tier1 test.

libs/forklift_inventory.py
```diff
@@ -8,9 +10,62 @@
...
+    if provider.ocp_resource is None:
+        raise ValueError(f"{provider.type} provider ocp_resource is not set")
+
+    provider_class = PROVIDER_INVENTORY_MAP.get(provider.type)
+    if provider_class is None:
+        raise ValueError(f"Provider {provider.type} not implemented")
+
+    # Subclasses use 'namespace' param while base class uses 'mtv_namespace'
+    return provider_class(  # type: ignore[call-arg]
+        client=client,
```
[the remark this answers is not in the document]
AmenB: Already justified — the comment on the line directly above explains the reason: `# Subclasses use 'namespace' param while base class uses 'mtv_namespace'`. This is the same pattern the original `conftest.py` code used before the refactor.

utilities/utils.py
```diff
@@ -273,6 +275,7 @@ def create_source_provider(
     destination_ocp_secret: Secret,
     insecure: bool,
     tmp_dir: pytest.TempPathFactory | None = None,
+    ca_cert_key: str = "cacert",
```
myakove, line 278: **[SUGGESTION]** `create_source_provider` has no docstring despite signature change (new `ca_cert_key` parameter).

Per "Docstring Format (MUST)" — since the signature was modified with a new externally-used parameter, adding a docstring with `Args`, `Returns`, and `Raises` sections would document the new `ca_cert_key` behavior (defaults to `"cacert"`, used for ca.crt secret field testing).

myakove (commented):
## Code Review

Found **1** issue(s) in this PR:

### :bulb: Suggestions (1)

| File | Line | Issue |
|------|------|-------|
| `utilities/utils.py` | 278 | **[SUGGESTION]** `create_source_provider` has no docstring despite signature cha |

---
*Review generated by pi*

rh-bot-1:
**Clean rebase detected** — no code changes compared to previous head (`7ec4f26`).

utilities/utils.py
AmenB: Fixed in 023a1df — added docstring with Args, Yields, and Raises sections to `create_source_provider`.

tests/cold/test_ca_crt_migration.py
```diff
@@ -0,0 +1,166 @@
...
+            fixture_store=fixture_store,
+            source_provider=ca_crt_source_provider,
+            destination_provider=destination_provider,
+            storage_map=self.storage_map,
+            network_map=self.network_map,
+            virtual_machines_list=prepared_plan["virtual_machines"],
+            target_namespace=target_namespace,
+            warm_migration=prepared_plan.get("warm_migration", False),
+        )
+        assert self.plan_resource, "Plan creation failed"
```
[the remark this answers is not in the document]
AmenB: Good point. Added `test_verify_ca_crt_secret` as the first test method — it directly asserts that the provider secret has `ca.crt` and does NOT have `cacert` before any migration steps run. Fixed in e3f1d98.

utilities/utils.py
AmenB: Already addressed — `create_source_provider` has a full docstring with Args (including `ca_cert_key`), Yields, and Raises sections. This was added in an earlier commit.

conftest.py
```diff
@@ -50,6 +50,7 @@
     OvaForkliftInventory,
     OvirtForkliftInventory,
     VsphereForkliftInventory,
+    create_forklift_inventory,
 )
```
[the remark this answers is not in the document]
AmenB: Fixed in 56b32bd — removed unused concrete `*ForkliftInventory` imports from `conftest.py`. Only `ForkliftInventory` and `create_forklift_inventory` remain.

tests/cold/test_ca_crt_migration.py
```diff
@@ -0,0 +1,186 @@
...
+    storage_map: StorageMap
+    network_map: NetworkMap
+    plan_resource: Plan
+
+    def test_verify_ca_crt_secret(
+        self,
+        ca_crt_source_provider: BaseProvider,
+        ocp_admin_client: "DynamicClient",
+        target_namespace: str,
+    ) -> None:
```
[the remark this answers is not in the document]
AmenB: Fixed in 56b32bd — removed unused `target_namespace` from `test_verify_ca_crt_secret`.

tests/cold/test_ca_crt_migration.py
krcmarik: Hey @AmenB, I looked into this more closely. The log message you're referencing ("CA certificate was not provided, system CA cert pool is used") is emitted in pkg/lib/client/openstack/client.go ->  getTLSConfig() -> Authenticate(), which gets called during provider reconciliation - the connection test and inventory collection. That all happens when the Provider CR is created, i.e. in your ca_crt_source_provider fixture, before any test method even runs.

So a 3-step test would produce the exact same controller logs. The migration doesn't add any additional CA cert code path - it just re-authenticates through the same Authenticate() -> getTLSConfig() chain.

Your test_verify_ca_crt_secret assertion is actually the most valuable thing here - it directly validates what we care about (secret has ca.crt, not cacert). Combined with the provider reaching Ready and Plan resolving VMs, that's full coverage of the feature. Our test environments mostly use self-signed CA certs that aren't in the system CA bundle, so if the secret only has ca.crt and the provider still reaches Ready - that's proof Forklift actually read and used ca.crt. If it fell back to the system pool, the TLS handshake would fail. For OpenStack specifically (where the CA might be in the system bundle), neither 3-step nor 5-step can prove it without log scraping anyway as you already mentioned.
If we skip original source VM cloning (since the creation of Provider/Plan is read-only operation) and we do not migrate any VMs We can have 1 minute test vs 20-30mins test with in my opinion identical test coverage. I'd keep it at 3 steps + the secret assertion.

AmenB: I see, i thought the issue occurred only when running the migration, therefore, i'll limit the tests to step 3 to avoid unnecessary run time
Thanks!

libs/forklift_inventory.py
```diff
@@ -8,9 +10,62 @@
...
+        Provider.ProviderType.OPENSHIFT: OpenshiftForkliftInventory,
+        Provider.ProviderType.OPENSTACK: OpenstackForliftinventory,
+    })
+
+
+def create_forklift_inventory(
+    client: DynamicClient,
+    mtv_namespace: str,
+    provider: "BaseProvider",
+) -> "ForkliftInventory":
```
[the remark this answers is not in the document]
AmenB: Not applicable — `DynamicClient` is already imported at runtime on line 4 of this file (`from kubernetes.dynamic.client import DynamicClient`). Every inventory class constructor uses it as a parameter type. Our function follows the same pre-existing pattern. The CLAUDE.md rule prohibits *instantiating* `DynamicClient(...)` directly (use `get_client()` instead), not using it as a type annotation — which is what we do here.

myakove (commented):
## Code Review

### :bulb: Suggestions (2)

| File | Line | Issue |
|------|------|-------|
| `libs/forklift_inventory.py` | 62 | **`type: ignore[call-arg]` without explanation** |
| `utilities/utils.py` | 282 | **[SUGGESTION]** Docstring Args missing type annotations per project standard. |

---
*Review generated by pi*

libs/forklift_inventory.py
```diff
@@ -8,9 +10,62 @@
...
+
+    if provider.ocp_resource is None:
+        raise ValueError(f"{provider.type} provider ocp_resource is not set")
+
+    provider_class = PROVIDER_INVENTORY_MAP.get(provider.type)
+    if provider_class is None:
+        raise ValueError(f"Provider {provider.type} not implemented")
+
+    # Subclasses use 'namespace' param while base class uses 'mtv_namespace'
+    return provider_class(  # type: ignore[call-arg]
```
myakove: **`type: ignore[call-arg]` without explanation**

Per "type: ignore without a comment explaining why" (No Dead Code or Unnecessary Comments — MUST).

**Fix:** Add an inline explanation:
```python
return provider_class(  # type: ignore[call-arg]  # Subclasses accept provider-specific kwargs
```

utilities/utils.py
```diff
@@ -273,7 +275,30 @@ def create_source_provider(
...
 ) -> Generator[BaseProvider, None, None]:
+    """Source provider configured and registered as a Forklift Provider CR.
+
+    Args:
```
myakove, line 282: **[SUGGESTION]** Docstring Args missing type annotations per project standard.

Args use bare names (`source_provider_data:`) instead of the AGENTS.md format (`source_provider_data (dict[str, Any]):`).

The existing docstrings in this codebase (e.g., `check_boot_configuration`, `verify_luks_encryption`) use the `param_name (Type): description` format.

**Fix:** Add type annotations to each arg:
```python
Args:
    source_provider_data (dict[str, Any]): Provider configuration from providers JSON.
    namespace (str): Target namespace...
```

AmenB: Fixed in 461e73a — added type annotations to all Args entries in `create_source_provider` docstring.

libs/forklift_inventory.py
AmenB: Fixed in 461e73a — moved the explanation inline: `# type: ignore[call-arg]  # Subclasses use 'namespace' param while base class uses 'mtv_namespace'`

tests/cold/conftest.py
```diff
@@ -0,0 +1,84 @@
...
+    with create_source_provider(
+        fixture_store=fixture_store,
+        session_uuid=session_uuid,
+        source_provider_data=source_provider_data,
+        namespace=target_namespace,
+        admin_client=ocp_admin_client,
+        tmp_dir=tmp_path_factory,
+        ocp_admin_client=ocp_admin_client,
+        destination_ocp_secret=destination_ocp_secret,
+        insecure=get_value_from_py_config(value="source_provider_insecure_skip_verify"),
```
myakove, line 53: **[SUGGESTION]** `insecure` receives `None` instead of `bool` when config key is absent.

`get_value_from_py_config(...)` returns `None` when the key doesn't exist, but `create_source_provider` declares `insecure: bool`. Works because `None` is falsy, but type mismatch.

```suggestion
        insecure=bool(get_value_from_py_config(value="source_provider_insecure_skip_verify")),
```

myakove (commented):
## Code Review

Found **1** issue(s) in this PR:

### :bulb: Suggestions (1)

| File | Line | Issue |
|------|------|-------|
| `tests/cold/conftest.py` | 53 | **[SUGGESTION]** `insecure` receives `None` instead of `bool` when config key is |

---
*Review generated by pi*

tests/cold/conftest.py
AmenB: Fixed in 1c60216 — wrapped with `bool()` to ensure `insecure` is always `bool`, not `None`.

tests/cold/test_ca_crt_migration.py
```diff
@@ -0,0 +1,152 @@
...
+    storage_map: StorageMap
+    network_map: NetworkMap
+    plan_resource: Plan
+
+    def test_verify_ca_crt_secret(
+        self,
+        prepared_plan: dict[str, Any],
+        ca_crt_source_provider: BaseProvider,
+        ocp_admin_client: "DynamicClient",
+    ) -> None:
```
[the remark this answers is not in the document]
AmenB: Cannot remove — `prepared_plan` is required here to resolve the `class_plan_config` indirect parametrize on the class. Without it, pytest fails at collection with: `function uses no fixture 'class_plan_config'`. We hit this exact error earlier and added `prepared_plan` to fix it.

tests/cold/test_ca_crt_migration.py
```diff
@@ -0,0 +1,152 @@
...
+
+    storage_map: StorageMap
+    network_map: NetworkMap
+    plan_resource: Plan
+
+    def test_verify_ca_crt_secret(
+        self,
+        prepared_plan: dict[str, Any],
+        ca_crt_source_provider: BaseProvider,
+        ocp_admin_client: "DynamicClient",
```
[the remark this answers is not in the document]
AmenB: Fixed in 0dea27b — removed redundant quotes from all type annotations.

tests/cold/conftest.py
```diff
@@ -0,0 +1,84 @@
...
+    from ocp_resources.secret import Secret
+
+
+@pytest.fixture(scope="class")  # Class-scoped: separate provider per test class with ca.crt in the secret
+def ca_crt_source_provider(
+    fixture_store: dict[str, Any],
+    session_uuid: str,
+    source_provider_data: dict[str, Any],
+    target_namespace: str,
+    ocp_admin_client: "DynamicClient",
```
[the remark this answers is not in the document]
AmenB: Fixed in 0dea27b — removed redundant quotes from all type annotations.

libs/forklift_inventory.py
```diff
@@ -8,9 +10,61 @@
...
+
+    provider_class = PROVIDER_INVENTORY_MAP.get(provider.type)
+    if provider_class is None:
+        raise ValueError(f"Provider {provider.type} not implemented")
+
+    return provider_class(  # type: ignore[call-arg]  # Subclasses use 'namespace' param while base class uses 'mtv_namespace'
+        client=client,
+        namespace=mtv_namespace,
+        provider_name=provider.ocp_resource.name,
+    )
```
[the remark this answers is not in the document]
AmenB: Fixed in 0dea27b — removed redundant quotes from `PROVIDER_INVENTORY_MAP` and `create_forklift_inventory` annotations.

tests/cold/test_ca_crt_migration.py
```diff
@@ -0,0 +1,152 @@
...
+        self.__class__.network_map = get_network_migration_map(
+            fixture_store=fixture_store,
+            source_provider=ca_crt_source_provider,
+            destination_provider=destination_provider,
+            source_provider_inventory=ca_crt_source_provider_inventory,
+            ocp_admin_client=ocp_admin_client,
+            target_namespace=target_namespace,
+            multus_network_name=multus_network_name,
+            vms=vms,
+        )
```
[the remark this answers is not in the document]
AmenB: Not an issue in practice — both providers point to the same backend. The `ca_crt_source_provider` waits for `Provider.Status.READY` which requires inventory to be loaded (`InventoryCreated` condition). By the time the test creates StorageMap/NetworkMap, the VM already exists in the ca.crt provider's inventory because it was synced during provider creation. We've run this test across vSphere, RHV, and OpenStack without hitting this race.

tests/cold/conftest.py
```diff
@@ -0,0 +1,84 @@
...
+        admin_client=ocp_admin_client,
+        tmp_dir=tmp_path_factory,
+        ocp_admin_client=ocp_admin_client,
+        destination_ocp_secret=destination_ocp_secret,
+        insecure=bool(get_value_from_py_config(value="source_provider_insecure_skip_verify")),
+        ca_cert_key="ca.crt",
+    ) as _source_provider:
+        yield _source_provider
+
+    _source_provider.disconnect()
```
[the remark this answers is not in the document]
AmenB: Not applicable — this follows the exact same pattern as the session-scoped `source_provider` fixture in `conftest.py:730-741`. The `disconnect()` call after the context manager is intentional and pre-existing in the codebase.

tests/cold/test_ca_crt_migration.py
```diff
@@ -0,0 +1,152 @@
...
+    from kubernetes.dynamic import DynamicClient
+
+    from libs.providers.openshift import OCPProvider
+
+
+@pytest.mark.vsphere
+@pytest.mark.rhv
+@pytest.mark.openstack
+@pytest.mark.esxi
+@pytest.mark.tier1
```
myakove, line 32: 🔴 **[CRITICAL] No collection-time skip for OCP/OVA providers**

The PR description claims "skipped for OpenShift/OVA via collection-time hook" but no such hook exists. The `@pytest.mark.vsphere`, `@pytest.mark.rhv`, `@pytest.mark.openstack`, `@pytest.mark.esxi` markers are label-only — they don't trigger automatic skipping.

If `-m tier1` runs with an OCP or OVA provider, `ca_crt_source_provider` will attempt CA cert operations that don't apply (OCP uses `destination_ocp_secret` directly; OVA has no TLS), leading to test failures.

**Fix:** Add a skip rule in `conftest.py`'s `pytest_collection_modifyitems`:
```python
# Skip CA cert tests for providers that don't use CA certificates
ca_cert_unsupported = (
    Provider.ProviderType.OPENSHIFT,
    Provider.ProviderType.OVA,
)
if source_provider_type in ca_cert_unsupported:
    ca_cert_skip = pytest.mark.skip(
        reason=f"{source_provider_type} does not use CA certificates"
    )
    for item in items:
        if "ca_crt" in item.nodeid.lower():
            item.add_marker(ca_cert_skip)
```

---

tests/cold/test_ca_crt_migration.py
```diff
@@ -0,0 +1,152 @@
...
+    so plan readiness proves the feature works.
+    """
+
+    storage_map: StorageMap
+    network_map: NetworkMap
+    plan_resource: Plan
+
+    def test_verify_ca_crt_secret(
+        self,
+        prepared_plan: dict[str, Any],
```
myakove: ⚠️ **`prepared_plan` requested but value unused — use `@pytest.mark.usefixtures`**

Per AGENTS.md "Fixture Request Patterns (MUST)": use `@pytest.mark.usefixtures("prepared_plan")` for side-effect fixtures whose return value isn't needed by the test body.

The fixture chain setup (class_plan_config → prepared_plan) IS needed for incremental test ordering, but since the VALUE isn't used here, switch to `usefixtures`:

```suggestion
    @pytest.mark.usefixtures("prepared_plan")
    def test_verify_ca_crt_secret(
        self,
        ca_crt_source_provider: BaseProvider,
        ocp_admin_client: DynamicClient,
```

---

libs/forklift_inventory.py
```diff
@@ -8,9 +10,61 @@
...
 
 
+def _register_inventory_classes() -> None:
+    """Populate PROVIDER_INVENTORY_MAP after all classes are defined."""
+    PROVIDER_INVENTORY_MAP.update({
+        Provider.ProviderType.OVA: OvaForkliftInventory,
+        Provider.ProviderType.RHV: OvirtForkliftInventory,
+        Provider.ProviderType.VSPHERE: VsphereForkliftInventory,
+        Provider.ProviderType.OPENSHIFT: OpenshiftForkliftInventory,
+        Provider.ProviderType.OPENSTACK: OpenstackForliftinventory,
```
myakove, line 28: ⚠️ **ESXi missing from `PROVIDER_INVENTORY_MAP` — `create_forklift_inventory()` crashes for ESXi providers**

`_register_inventory_classes()` maps OVA, RHV, VSPHERE, OPENSHIFT, OPENSTACK but not ESXI. Since ESXi is essentially vSphere from a Forklift inventory perspective, it should map to `VsphereForkliftInventory`.

This affects BOTH:
1. The new `ca_crt_source_provider_inventory` fixture (test has `@pytest.mark.esxi`)
2. The existing session-scoped `source_provider_inventory` (refactored in this PR to use `create_forklift_inventory()`)

```suggestion
    PROVIDER_INVENTORY_MAP.update({
        Provider.ProviderType.OVA: OvaForkliftInventory,
        Provider.ProviderType.RHV: OvirtForkliftInventory,
        Provider.ProviderType.VSPHERE: VsphereForkliftInventory,
        Provider.ProviderType.OPENSHIFT: OpenshiftForkliftInventory,
        Provider.ProviderType.OPENSTACK: OpenstackForliftinventory,
        Provider.ProviderType.ESXI: VsphereForkliftInventory,
    })
```

---

myakove (commented):
## Code Review

### :bulb: Suggestions (3)

| File | Line | Issue |
|------|------|-------|
| `tests/cold/test_ca_crt_migration.py` | 31 | 🔴 **[CRITICAL] No collection-time skip for OCP/OVA providers** |
| `tests/cold/test_ca_crt_migration.py` | 62 | ⚠️ **`prepared_plan` requested but value unused — use `@pytest.mark.usefixtures` |
| `libs/forklift_inventory.py` | 28 | ⚠️ **ESXi missing from `PROVIDER_INVENTORY_MAP` — `create_forklift_inventory()`  |

---
*Review generated by pi*

---

libs/forklift_inventory.py
AmenB: `Provider.ProviderType.ESXI` does not exist — ESXi providers use `Provider.ProviderType.VSPHERE` as their type. Verified with `uv run python3 -c "from ocp_resources.provider import Provider; print(dir(Provider.ProviderType))"` — only `VSPHERE` exists, no `ESXI`. The `@pytest.mark.esxi` marker is for test filtering only, not a distinct provider type. The existing session-scoped `source_provider_inventory` also had no ESXI entry for the same reason.

tests/cold/test_ca_crt_migration.py
AmenB: Fixed in f07141c — switched to `@pytest.mark.usefixtures("prepared_plan")` since the value isn't used in the test body.

tests/cold/test_ca_crt_migration.py
AmenB: Fixed in f07141c — added collection-time skip in `pytest_collection_modifyitems` for OCP/OVA providers. Tests with `ca_crt` in the nodeid are skipped when the source provider is OpenShift or OVA.

conftest.py
```diff
@@ -334,6 +330,19 @@ def pytest_collection_modifyitems(session, config, items):
...
+            ca_cert_unsupported = (
+                Provider.ProviderType.OPENSHIFT,
+                Provider.ProviderType.OVA,
+            )
+            if source_provider_type in ca_cert_unsupported:
+                ca_cert_skip = pytest.mark.skip(
+                    reason=f"{source_provider_type} does not use CA certificates in provider secrets"
+                )
+                for item in items:
+                    if "ca_crt" in item.nodeid.lower():
```
myakove: **[WARNING] Collection-time skip uses `item.nodeid.lower()` instead of the documented marker-based pattern**

All other skip rules in this hook use `item.keywords` (marker-based matching). The documented 3-step pattern (lines 288–296) explicitly says:
1. Register marker in `pytest.ini` under `[pytest] > markers`
2. Use `item.keywords` matching
3. Apply marker to test class

The current `"ca_crt" in item.nodeid.lower()` is fragile — it matches ANY test with `ca_crt` anywhere in its file path/name. No `ca_crt` marker is registered in `pytest.ini`.

**Fix:**
1. Add to `pytest.ini` markers: `ca_crt: CA certificate field migration tests`
2. Add `@pytest.mark.ca_crt` to `TestCaCrtColdMigration`
3. Change this line:

```suggestion
                    if "ca_crt" in item.keywords:
```

---

myakove (commented):
## Code Review

Found **1** issue(s) in this PR:

### :bulb: Suggestions (1)

| File | Line | Issue |
|------|------|-------|
| `conftest.py` | 343 | **[WARNING] Collection-time skip uses `item.nodeid.lower()` instead of the docum |

---
*Review generated by pi*

---

conftest.py
AmenB: Fixed in 6ce5796 — followed the documented 3-step pattern:
1. Registered `ca_crt` marker in `pytest.ini`
2. Added `@pytest.mark.ca_crt` to `TestCaCrtColdMigration`
3. Changed to `item.keywords` matching in collection hook

tests/cold/test_ca_crt_migration.py
```diff
@@ -0,0 +1,153 @@
...
+    from libs.providers.openshift import OCPProvider
+
+
+@pytest.mark.vsphere
+@pytest.mark.rhv
+@pytest.mark.openstack
+@pytest.mark.esxi
+@pytest.mark.ca_crt
+@pytest.mark.tier1
+@pytest.mark.incremental
```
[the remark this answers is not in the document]
AmenB: This is a pre-existing behavior in the upstream code, not introduced by this PR. The `"vsphere" in item.keywords` skip logic on line 330 exists in upstream/main and affects ALL tests with `@pytest.mark.vsphere`, including `TestSanityColdMtvMigration` which has the same marker set. Our test follows the same pattern. If this needs fixing, it should be a separate PR addressing the upstream skip logic.

tests/cold/conftest.py
```diff
@@ -0,0 +1,84 @@
...
+        ocp_admin_client (DynamicClient): OpenShift admin client.
+        mtv_namespace (str): MTV operator namespace.
+        ca_crt_source_provider (BaseProvider): Source provider using ca.crt secret field.
+
+    Returns:
+        ForkliftInventory: Inventory instance for the ca.crt provider.
+
+    Raises:
+        ValueError: If ocp_resource is not set on the provider.
+    """
```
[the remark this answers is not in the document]
AmenB: Fixed in 0665b66 — removed `Raises` section from the wrapper fixture docstring. Exception documentation belongs in `create_forklift_inventory()`, not in the wrapper.

utilities/utils.py
```diff
@@ -160,21 +160,23 @@ def _fetch_and_store_cacert(
...
     Returns:
-        Path to the certificate file
+        Path: Path to the certificate file.
 
     Raises:
-        ValueError: If tmp_dir is not provided
+        ValueError: If tmp_dir is not provided.
     """
     source_provider_type = source_provider_data["type"]
     if not tmp_dir:
```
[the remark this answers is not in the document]
AmenB: The in-place mutation of `secret_string_data` is pre-existing behavior — the function name `_fetch_and_store_cacert` explicitly indicates it stores (mutates) the data. Our PR only added the `ca_cert_key` parameter; the mutation pattern was not introduced here. Documenting pre-existing mutation patterns is out of scope for this PR.

pytest.ini
```diff
@@ -30,6 +30,7 @@ markers =
     rhv: RHV provider-specific tests
     openstack: OpenStack provider-specific tests
     openshift: OpenShift provider-specific tests
+    ca_crt: CA certificate field (ca.crt) provider secret tests
```
myakove, line 34: **[CRITICAL] AGENTS.md Test Markers table not updated with new `ca_crt` marker**

New marker `ca_crt` added here but not documented in AGENTS.md's Test Markers table (around line 961). Per "(MUST) Update CLAUDE.md when methodology or coding patterns change."

Add to the markers table:
```
| `ca_crt`                | CA certificate field (ca.crt) tests    |
```

---

conftest.py
```diff
@@ -334,6 +330,19 @@ def pytest_collection_modifyitems(session, config, items):
...
+            ca_cert_unsupported = (
+                Provider.ProviderType.OPENSHIFT,
+                Provider.ProviderType.OVA,
+            )
+            if source_provider_type in ca_cert_unsupported:
+                ca_cert_skip = pytest.mark.skip(
+                    reason=f"{source_provider_type} does not use CA certificates in provider secrets"
+                )
+                for item in items:
+                    if "ca_crt" in item.keywords:
```
myakove, line 348: **[CRITICAL] AGENTS.md "Marker requirements for collection-time skipping" section not updated**

New collection-time skip rule for `ca_crt` (OCP/OVA providers) added here but not documented in AGENTS.md alongside `warm` and `copyoffload_snapshots` (around line 972).

Per "Marker requirements for collection-time skipping (MUST)" — add:
```
- CA certificate tests → `@pytest.mark.ca_crt`
```

Also update README.md's marker table to include `ca_crt`.

---

myakove (commented):
## Code Review

### :bulb: Suggestions (2)

| File | Line | Issue |
|------|------|-------|
| `pytest.ini` | 33 | **[CRITICAL] AGENTS.md Test Markers table not updated with new `ca_crt` marker** |
| `conftest.py` | 343 | **[CRITICAL] AGENTS.md "Marker requirements for collection-time skipping" sectio |

---
*Review generated by pi*

---

conftest.py
AmenB: Fixed in a0802a8 — added `ca_crt` to AGENTS.md's collection-time skip requirements section and README.md's marker table.

pytest.ini
AmenB: Fixed in a0802a8 — added `ca_crt` to AGENTS.md's Test Markers table and README.md's marker table.

myakove (commented):
## Code Review

Found **1** issue(s) in this PR:

### :bulb: Suggestions (1)

| File | Line | Issue |
|------|------|-------|
| `tests/cold/conftest.py` | 53 | 💡 **Inconsistent `bool()` wrapping of `get_value_from_py_config()`** |

---
*Review generated by pi*

---

tests/cold/conftest.py
```diff
@@ -0,0 +1,81 @@
...
+    with create_source_provider(
+        fixture_store=fixture_store,
+        session_uuid=session_uuid,
+        source_provider_data=source_provider_data,
+        namespace=target_namespace,
+        admin_client=ocp_admin_client,
+        tmp_dir=tmp_path_factory,
+        ocp_admin_client=ocp_admin_client,
+        destination_ocp_secret=destination_ocp_secret,
+        insecure=bool(get_value_from_py_config(value="source_provider_insecure_skip_verify")),
```
myakove: 💡 **Inconsistent `bool()` wrapping of `get_value_from_py_config()`**

The session-scoped `source_provider` fixture (`conftest.py:747`) passes the value directly:
```python
insecure=get_value_from_py_config(value="source_provider_insecure_skip_verify"),
```

This new fixture wraps with `bool()`. Both work, but the inconsistency could cause confusion if `create_source_provider` ever adds strict type checking.

```suggestion
        insecure=get_value_from_py_config(value="source_provider_insecure_skip_verify"),
```

---

AmenB: Fixed in 7649725 — removed `bool()` to match the session-scoped `source_provider` fixture pattern in `conftest.py:747`.

tests/cold/test_ca_crt_migration.py
```diff
@@ -0,0 +1,153 @@
...
+        """Verify the provider secret uses ca.crt field instead of cacert."""
+        assert ca_crt_source_provider.ocp_resource is not None, "ocp_resource is not set"
+        secret_ref = ca_crt_source_provider.ocp_resource.instance.spec.secret
+        secret = Secret(
+            client=ocp_admin_client,
+            name=secret_ref.name,
+            namespace=secret_ref.namespace,
+        )
+        secret_data_keys = list(secret.instance.data.keys())
+        assert "ca.crt" in secret_data_keys, f"Expected 'ca.crt' in secret keys, got: {secret_data_keys}"
```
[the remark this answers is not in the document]
AmenB: Not an issue — the secret is created by `create_source_provider()` which waits for the secret to exist before creating the Provider CR. The provider then reaches `Ready` status (which validates the secret). By the time this test runs, the secret has been fully populated for several minutes. A race condition is not possible here.

libs/forklift_inventory.py
```diff
@@ -8,9 +10,61 @@
 from simple_logger.logger import get_logger
 from timeout_sampler import TimeoutExpiredError, TimeoutSampler
 
+if TYPE_CHECKING:
```
myakove, line 13: ⚠️ **`DynamicClient` runtime import should move under `TYPE_CHECKING`**

Line 6 imports `DynamicClient` at runtime from `kubernetes.dynamic.client`, but it's only used in type annotations (no `isinstance`, no runtime usage). Per AGENTS.md DynamicClient Rules: runtime import from `kubernetes.*` is forbidden; `TYPE_CHECKING` block import is allowed.

This PR already adds `from __future__ import annotations` and a `TYPE_CHECKING` block — the move is trivial:

```python
if TYPE_CHECKING:
    from kubernetes.dynamic import DynamicClient

    from libs.base_provider import BaseProvider
```

And remove line 6: `from kubernetes.dynamic.client import DynamicClient`

---

AGENTS.md
```diff
@@ -973,6 +973,7 @@ Class-scoped teardown fixture that cleans up migrated VMs after each test class
 | `copyoffload`           | Copy-offload (XCOPY) tests             |
 | `copyoffload_sanity`    | Copy-offload sanity subset             |
 | `copyoffload_snapshots` | Copy-offload snapshot tests (vSphere)  |
+| `ca_crt`                | CA certificate field (ca.crt) tests    |
```
myakove, line 994: 💡 **New 4-step plan-readiness test pattern undocumented**

`TestCaCrtColdMigration` introduces a new pattern: `verify_<feature>` → `storagemap` → `networkmap` → `plan` — intentionally skipping migration and VM checks. AGENTS.md documents 5/6/7-step patterns but not this subset.

Suggestion — add under Key Patterns:
```
- **4-step plan-readiness pattern**: verify_<feature> -> storagemap -> networkmap -> plan
  Used when the feature under test is exercised during provider/plan creation (e.g., CA cert field
  validation). No migration is executed — plan readiness proves the feature works.
```

---

myakove (commented):
## Code Review

### :bulb: Suggestions (2)

| File | Line | Issue |
|------|------|-------|
| `libs/forklift_inventory.py` | 13 | ⚠️ **`DynamicClient` runtime import should move under `TYPE_CHECKING`** |
| `AGENTS.md` | 976 | 💡 **New 4-step plan-readiness test pattern undocumented** |

---
*Review generated by pi*

---

AGENTS.md
AmenB: Fixed in 92a30ca — added 4-step plan-readiness pattern to AGENTS.md Key Patterns section, before the 5-step pattern.

libs/forklift_inventory.py
```diff
@@ -8,9 +10,61 @@
...
+        ValueError: If the provider type is not supported.
+    """
+    if not PROVIDER_INVENTORY_MAP:
+        _register_inventory_classes()
+
+    if provider.ocp_resource is None:
+        raise ValueError(f"{provider.type} provider ocp_resource is not set")
+
+    provider_class = PROVIDER_INVENTORY_MAP.get(provider.type)
+    if provider_class is None:
```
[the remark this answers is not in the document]
AmenB: `Provider.ProviderType` values are string constants (e.g., `"vsphere"`, `"ovirt"`, `"openstack"`). The `dict[str, ...]` annotation is correct — keys are strings both in the annotation and at runtime. No type inconsistency.

rh-bot-1:

PR #566, opened 2026-06-19
feat: add firmware, secure boot, and TPM verification to post-migration checks

krcmarik:
Add firmware metadata extraction and verification to ensure migration preserves boot configuration:

- VMware provider: extract boot firmware type (bios/efi), secure boot setting, and TPM presence from source VMs
- OpenShift provider: extract the same firmware metadata from destination CNV VMs
- post_migration: add check_boot_configuration() that asserts source and destination firmware type, secure boot, and TPM all match
- Wire into existing check_vms() flow so vSphere migration tests automatically validate firmware preservation

## Summary by CodeRabbit

* **New Features**
  * VM details now include firmware information for supported providers: boot firmware type (EFI/BIOS), secure boot status, TPM presence, and serial when available.
* **Bug Fixes**
  * Improved and standardized firmware/serial extraction logic, with stricter validation for EFI secure boot/boot options where required.
* **Tests / Validation**
  * Enhanced post-migration checks to compare boot firmware, secure boot, and TPM settings between source and destination, with clearer per-VM mismatch reporting.

### DIFF
--- a/libs/providers/openshift.py
+++ b/libs/providers/openshift.py
@@ -200,7 +200,7 @@ class OCPProvider(BaseProvider):
 
         devices: dict[str, Any] = domain.get("devices", {})
         tpm_config: dict[str, Any] | None = devices.get("tpm")
-        firmware_info["tpm_present"] = tpm_config is not None
+        firmware_info["tpm_present"] = bool(tpm_config.get("persistent")) if tpm_config else False
 
         bootloader: dict[str, Any] | None = firmware_spec.get("bootloader") if firmware_spec else None
         efi_config: dict[str, Any] | None = bootloader.get("efi") if bootloader else None
--- a/libs/providers/openshift.py
+++ b/libs/providers/openshift.py
@@ -122,7 +122,8 @@ class OCPProvider(BaseProvider):
             dict[str, Any]: VM information dictionary.
 
         Raises:
-            ValueError: If `ocp_resource` is not set or VM fails to start.
+            ValueError: If `ocp_resource` is not set, VM fails to start,
+                or secureBoot is missing from an EFI VM's firmware config.
             InvalidVMNameError: If destination VM name fails Kubernetes validation.
         """
         if not self.ocp_resource:
--- a/libs/providers/openshift.py
+++ b/libs/providers/openshift.py
@@ -192,8 +192,27 @@ class OCPProvider(BaseProvider):
         )
 
         # Serial number (from firmware) - for serial preservation verification
-        firmware_spec: dict[str, Any] | None = cnv_vm.instance.spec.template.spec.domain.get("firmware")
+        domain: dict[str, Any] = cnv_vm.instance.spec.template.spec.domain
+
+        firmware_spec: dict[str, Any] | None = domain.get("firmware")
         result_vm_info["serial"] = firmware_spec.get("serial") if firmware_spec else None
+        firmware_info: dict[str, Any] = {}
+
+        devices: dict[str, Any] = domain.get("devices", {})
+        tpm_config: dict[str, Any] | None = devices.get("tpm")
+        firmware_info["tpm_present"] = bool(tpm_config.get("persistent")) if tpm_config else False
+
+        bootloader = firmware_spec.get("bootloader") if firmware_spec else None
+        efi_config = bootloader.get("efi") if bootloader else None
+
+        if efi_config:
+            firmware_info["boot_firmware"] = "efi"
+            firmware_info["secure_boot"] = efi_config["secureBoot"]
+        else:
+            firmware_info["boot_firmware"] = "bios"
+            firmware_info["secure_boot"] = False
+
+        result_vm_info["firmware"] = firmware_info
 
         # Extract template once to avoid duplicate attribute access
         template = cnv_vm.instance.spec.template
--- a/libs/providers/openshift.py
+++ b/libs/providers/openshift.py
@@ -196,10 +196,12 @@ class OCPProvider(BaseProvider):
 
         firmware_spec: dict[str, Any] | None = domain.get("firmware")
         result_vm_info["serial"] = firmware_spec.get("serial") if firmware_spec else None
+        # Firmware configuration — boot type, secure boot, TPM
         firmware_info: dict[str, Any] = {}
 
         devices: dict[str, Any] = domain.get("devices", {})
         tpm_config: dict[str, Any] | None = devices.get("tpm")
+        # KubeVirt adds empty tpm: {} even for VMs without TPM; only persistent=true indicates actual TPM
         firmware_info["tpm_present"] = bool(tpm_config.get("persistent")) if tpm_config else False
 
         bootloader: dict[str, Any] | None = firmware_spec.get("bootloader") if firmware_spec else None
--- a/libs/providers/openshift.py
+++ b/libs/providers/openshift.py
@@ -200,14 +200,17 @@ class OCPProvider(BaseProvider):
 
         devices: dict[str, Any] = domain.get("devices", {})
         tpm_config: dict[str, Any] | None = devices.get("tpm")
-        firmware_info["tpm_present"] = bool(tpm_config.get("persistent")) if tpm_config else False
+        firmware_info["tpm_present"] = tpm_config is not None
 
-        bootloader = firmware_spec.get("bootloader") if firmware_spec else None
-        efi_config = bootloader.get("efi") if bootloader else None
+        bootloader: dict[str, Any] | None = firmware_spec.get("bootloader") if firmware_spec else None
+        efi_config: dict[str, Any] | None = bootloader.get("efi") if bootloader else None
 
         if efi_config:
             firmware_info["boot_firmware"] = "efi"
-            firmware_info["secure_boot"] = efi_config["secureBoot"]
+            secure_boot: bool | None = efi_config.get("secureBoot")
+            if secure_boot is None:
+                raise ValueError(f"secureBoot not found in EFI config for VM '{cnv_vm_name}'")
+            firmware_info["secure_boot"] = secure_boot
         else:
             firmware_info["boot_firmware"] = "bios"
             firmware_info["secure_boot"] = False
--- a/libs/providers/vmware.py
+++ b/libs/providers/vmware.py
@@ -735,6 +735,8 @@ class VMWareProvider(BaseProvider):
         if not boot_firmware_raw:
             raise ValueError(f"firmware attribute is missing or empty for VM '{_vm.name}'")
         boot_firmware: str = boot_firmware_raw.lower()
+        if boot_firmware not in ("bios", "efi"):
+            raise ValueError(f"Unexpected firmware type '{boot_firmware}' for VM '{_vm.name}'")
         firmware_info["boot_firmware"] = boot_firmware
         boot_options: vim.vm.BootOptions | None = vm_config.bootOptions
         if boot_firmware == "efi":
--- a/libs/providers/vmware.py
+++ b/libs/providers/vmware.py
@@ -736,10 +736,9 @@ class VMWareProvider(BaseProvider):
         boot_firmware: str = boot_firmware_raw.lower()
         firmware_info["boot_firmware"] = boot_firmware
         boot_options: vim.vm.BootOptions | None = vm_config.bootOptions
-        if boot_firmware == "efi" and boot_options is None:
-            raise ValueError(f"bootOptions is None for EFI VM '{_vm.name}' — expected boot configuration")
         if boot_firmware == "efi":
-            assert boot_options is not None  # Guaranteed by ValueError check above
+            if boot_options is None:
+                raise ValueError(f"bootOptions is None for EFI VM '{_vm.name}' — expected boot configuration")
             firmware_info["secure_boot"] = boot_options.efiSecureBootEnabled
         else:
             firmware_info["secure_boot"] = False
--- a/libs/providers/vmware.py
+++ b/libs/providers/vmware.py
@@ -730,8 +730,11 @@ class VMWareProvider(BaseProvider):
         result_vm_info["win_os"] = "win" in vm_config.guestId
 
         firmware_info: dict[str, Any] = {}
-        firmware_info["boot_firmware"] = vm_config.firmware
+        boot_firmware = vm_config.firmware.lower()
+        firmware_info["boot_firmware"] = boot_firmware
         boot_options = vm_config.bootOptions
+        if boot_firmware == "efi" and boot_options is None:
+            raise ValueError(f"bootOptions is None for EFI VM '{_vm.name}' — expected boot configuration")
         firmware_info["secure_boot"] = boot_options.efiSecureBootEnabled if boot_options else False
 
         firmware_info["tpm_present"] = any(
--- a/libs/providers/vmware.py
+++ b/libs/providers/vmware.py
@@ -730,9 +730,12 @@ class VMWareProvider(BaseProvider):
         result_vm_info["win_os"] = "win" in vm_config.guestId
 
         firmware_info: dict[str, Any] = {}
-        boot_firmware = vm_config.firmware.lower()
+        boot_firmware_raw: str | None = vm_config.firmware
+        if not boot_firmware_raw:
+            raise ValueError(f"firmware attribute is missing or empty for VM '{_vm.name}'")
+        boot_firmware: str = boot_firmware_raw.lower()
         firmware_info["boot_firmware"] = boot_firmware
-        boot_options = vm_config.bootOptions
+        boot_options: vim.vm.BootOptions | None = vm_config.bootOptions
         if boot_firmware == "efi" and boot_options is None:
             raise ValueError(f"bootOptions is None for EFI VM '{_vm.name}' — expected boot configuration")
         firmware_info["secure_boot"] = boot_options.efiSecureBootEnabled if boot_options else False
--- a/libs/providers/vmware.py
+++ b/libs/providers/vmware.py
@@ -729,6 +729,16 @@ class VMWareProvider(BaseProvider):
         # Guest OS
         result_vm_info["win_os"] = "win" in vm_config.guestId
 
+        firmware_info: dict[str, Any] = {}
+        firmware_info["boot_firmware"] = vm_config.firmware
+        boot_options = vm_config.bootOptions
+        firmware_info["secure_boot"] = boot_options.efiSecureBootEnabled if boot_options else False
+
+        firmware_info["tpm_present"] = any(
+            isinstance(device, vim.vm.device.VirtualTPM) for device in vm_config.hardware.device
+        )
+        result_vm_info["firmware"] = firmware_info
+
         # Power state
         if _vm.runtime.powerState == "poweredOn":
             result_vm_info["power_state"] = "on"
--- a/libs/providers/vmware.py
+++ b/libs/providers/vmware.py
@@ -738,7 +738,11 @@ class VMWareProvider(BaseProvider):
         boot_options: vim.vm.BootOptions | None = vm_config.bootOptions
         if boot_firmware == "efi" and boot_options is None:
             raise ValueError(f"bootOptions is None for EFI VM '{_vm.name}' — expected boot configuration")
-        firmware_info["secure_boot"] = boot_options.efiSecureBootEnabled if boot_options else False
+        if boot_firmware == "efi":
+            assert boot_options is not None  # Guaranteed by ValueError check above
+            firmware_info["secure_boot"] = boot_options.efiSecureBootEnabled
+        else:
+            firmware_info["secure_boot"] = False
 
         firmware_info["tpm_present"] = any(
             isinstance(device, vim.vm.device.VirtualTPM) for device in vm_config.hardware.device
--- a/libs/providers/vmware.py
+++ b/libs/providers/vmware.py
@@ -729,6 +729,7 @@ class VMWareProvider(BaseProvider):
         # Guest OS
         result_vm_info["win_os"] = "win" in vm_config.guestId
 
+        # Firmware configuration — boot type, secure boot, TPM
         firmware_info: dict[str, Any] = {}
         boot_firmware_raw: str | None = vm_config.firmware
         if not boot_firmware_raw:
--- a/utilities/post_migration.py
+++ b/utilities/post_migration.py
@@ -1266,6 +1266,7 @@ def check_boot_configuration(source_vm: dict[str, Any], destination_vm: dict[str
 
     Raises:
         AssertionError: If firmware type, secure boot, or TPM settings don't match.
+        KeyError: If required firmware keys are missing in source or destination VM info.
     """
     source_firmware: dict[str, Any] = source_vm["firmware"]
     dest_firmware: dict[str, Any] = destination_vm["firmware"]
@@ -1668,7 +1669,7 @@ def check_vms(
 
             try:
                 check_boot_configuration(source_vm=source_vm, destination_vm=destination_vm)
-            except (AssertionError, KeyError) as exp:
+            except Exception as exp:
                 res[vm_name].append(f"check_boot_configuration - {str(exp)}")
 
         # Group 4: RHV-specific checks
--- a/utilities/post_migration.py
+++ b/utilities/post_migration.py
@@ -1276,11 +1276,12 @@ def check_boot_configuration(source_vm: dict[str, Any], destination_vm: dict[str
         ("secure_boot", "Secure boot"),
         ("tpm_present", "TPM"),
     ]
+    mismatches: list[str] = []
     for key, label in firmware_checks:
-        assert source_firmware[key] == dest_firmware[key], (
-            f"{label} mismatch for VM '{destination_vm['name']}': "
-            f"source={source_firmware[key]}, destination={dest_firmware[key]}"
-        )
+        if source_firmware[key] != dest_firmware[key]:
+            mismatches.append(f"{label}: source={source_firmware[key]}, destination={dest_firmware[key]}")
+    if mismatches:
+        raise AssertionError(f"Firmware mismatch for VM '{destination_vm['name']}': {'; '.join(mismatches)}")
 
     LOGGER.info(
         f"Firmware checks passed for VM '{destination_vm['name']}': "
--- a/utilities/post_migration.py
+++ b/utilities/post_migration.py
@@ -1257,8 +1257,8 @@ def check_serial_preservation(
         LOGGER.info(f"Serial preserved correctly (OCP {ocp_version} < 4.20, plain UUID): {dest_serial}")
 
 
-def check_firmware_and_tpm(source_vm: dict[str, Any], destination_vm: dict[str, Any]) -> None:
-    """Verify firmware type and TPM settings are preserved after migration.
+def check_boot_configuration(source_vm: dict[str, Any], destination_vm: dict[str, Any]) -> None:
+    """Verify boot configuration is preserved after migration.
 
     Args:
         source_vm: Source VM info dictionary with firmware data.
@@ -1270,7 +1270,11 @@ def check_firmware_and_tpm(source_vm: dict[str, Any], destination_vm: dict[str, 
     source_firmware: dict[str, Any] = source_vm["firmware"]
     dest_firmware: dict[str, Any] = destination_vm["firmware"]
 
-    firmware_checks = [("boot_firmware", "Boot firmware"), ("secure_boot", "Secure boot"), ("tpm_present", "TPM")]
+    firmware_checks: list[tuple[str, str]] = [
+        ("boot_firmware", "Boot firmware"),
+        ("secure_boot", "Secure boot"),
+        ("tpm_present", "TPM"),
+    ]
     for key, label in firmware_checks:
         assert source_firmware[key] == dest_firmware[key], (
             f"{label} mismatch for VM '{destination_vm['name']}': "
@@ -1663,9 +1667,9 @@ def check_vms(
                 res[vm_name].append(f"check_serial_preservation - {str(exp)}")
 
             try:
-                check_firmware_and_tpm(source_vm=source_vm, destination_vm=destination_vm)
-            except Exception as exp:
-                res[vm_name].append(f"check_firmware_and_tpm - {str(exp)}")
+                check_boot_configuration(source_vm=source_vm, destination_vm=destination_vm)
+            except (AssertionError, KeyError) as exp:
+                res[vm_name].append(f"check_boot_configuration - {str(exp)}")
 
         # Group 4: RHV-specific checks
         if rhv_provider(source_provider_data) and isinstance(source_provider, OvirtProvider):
--- a/utilities/post_migration.py
+++ b/utilities/post_migration.py
@@ -1170,6 +1170,45 @@ def check_serial_preservation(
         LOGGER.info(f"Serial preserved correctly (OCP {ocp_version} < 4.20, plain UUID): {dest_serial}")
 
 
+def check_firmware_and_tpm(source_vm: dict[str, Any], destination_vm: dict[str, Any]) -> None:
+    """Verify firmware type and TPM settings are preserved after migration.
+
+    Args:
+        source_vm: Source VM info dictionary with firmware data.
+        destination_vm: Destination VM info dictionary with firmware data.
+
+    Raises:
+        AssertionError: If firmware type, secure boot, or TPM settings don't match.
+    """
+    source_firmware: dict[str, Any] = source_vm["firmware"]
+    dest_firmware: dict[str, Any] = destination_vm["firmware"]
+
+    source_boot = source_firmware["boot_firmware"]
+    dest_boot = dest_firmware["boot_firmware"]
+    assert source_boot == dest_boot, (
+        f"Boot firmware mismatch for VM '{destination_vm['name']}': source={source_boot}, destination={dest_boot}"
+    )
+
+    source_secure_boot = source_firmware["secure_boot"]
+    dest_secure_boot = dest_firmware["secure_boot"]
+    assert source_secure_boot == dest_secure_boot, (
+        f"Secure boot mismatch for VM '{destination_vm['name']}': "
+        f"source={source_secure_boot}, destination={dest_secure_boot}"
+    )
+
+    source_tpm = source_firmware["tpm_present"]
+    dest_tpm = dest_firmware["tpm_present"]
+    assert source_tpm == dest_tpm, (
+        f"TPM mismatch for VM '{destination_vm['name']}': "
+        f"source tpm_present={source_tpm}, destination tpm_present={dest_tpm}"
+    )
+
+    LOGGER.info(
+        f"Firmware checks passed for VM '{destination_vm['name']}': "
+        f"boot={dest_boot}, secure_boot={dest_secure_boot}, tpm={dest_tpm}"
+    )
+
+
 def check_vm_node_placement(
     destination_vm: dict[str, Any],
     expected_node: str,
@@ -1547,6 +1586,11 @@ def check_vms(
                 )
             except Exception as exp:
                 res[vm_name].append(f"check_serial_preservation - {str(exp)}")
+
+            try:
+                check_firmware_and_tpm(source_vm=source_vm, destination_vm=destination_vm)
+            except Exception as exp:
+                res[vm_name].append(f"check_firmware_and_tpm - {str(exp)}")
 
         # Group 4: RHV-specific checks
         if rhv_provider(source_provider_data) and isinstance(source_provider, OvirtProvider):
--- a/utilities/post_migration.py
+++ b/utilities/post_migration.py
@@ -1183,29 +1183,17 @@ def check_firmware_and_tpm(source_vm: dict[str, Any], destination_vm: dict[str, 
     source_firmware: dict[str, Any] = source_vm["firmware"]
     dest_firmware: dict[str, Any] = destination_vm["firmware"]
 
-    source_boot = source_firmware["boot_firmware"]
-    dest_boot = dest_firmware["boot_firmware"]
-    assert source_boot == dest_boot, (
-        f"Boot firmware mismatch for VM '{destination_vm['name']}': source={source_boot}, destination={dest_boot}"
-    )
-
-    source_secure_boot = source_firmware["secure_boot"]
-    dest_secure_boot = dest_firmware["secure_boot"]
-    assert source_secure_boot == dest_secure_boot, (
-        f"Secure boot mismatch for VM '{destination_vm['name']}': "
-        f"source={source_secure_boot}, destination={dest_secure_boot}"
-    )
-
-    source_tpm = source_firmware["tpm_present"]
-    dest_tpm = dest_firmware["tpm_present"]
-    assert source_tpm == dest_tpm, (
-        f"TPM mismatch for VM '{destination_vm['name']}': "
-        f"source tpm_present={source_tpm}, destination tpm_present={dest_tpm}"
-    )
+    firmware_checks = [("boot_firmware", "Boot firmware"), ("secure_boot", "Secure boot"), ("tpm_present", "TPM")]
+    for key, label in firmware_checks:
+        assert source_firmware[key] == dest_firmware[key], (
+            f"{label} mismatch for VM '{destination_vm['name']}': "
+            f"source={source_firmware[key]}, destination={dest_firmware[key]}"
+        )
 
     LOGGER.info(
         f"Firmware checks passed for VM '{destination_vm['name']}': "
-        f"boot={dest_boot}, secure_boot={dest_secure_boot}, tpm={dest_tpm}"
+        f"boot={dest_firmware['boot_firmware']}, secure_boot={dest_firmware['secure_boot']}, "
+        f"tpm={dest_firmware['tpm_present']}"
     )
 
 
--- a/libs/base_provider.py
+++ b/libs/base_provider.py
@@ -26,6 +26,7 @@ class BaseProvider(abc.ABC):
         "memory_in_mb": 0,
         "snapshots_data": [],
         "power_state": "",
+        "firmware": {},
     }
 
     def __init__(


utilities/post_migration.py
```diff
@@ -1170,6 +1170,45 @@ def check_serial_preservation(
...
+def check_firmware_and_tpm(source_vm: dict[str, Any], destination_vm: dict[str, Any]) -> None:
+    """Verify firmware type and TPM settings are preserved after migration.
+
+    Args:
+        source_vm: Source VM info dictionary with firmware data.
+        destination_vm: Destination VM info dictionary with firmware data.
+
+    Raises:
+        AssertionError: If firmware type, secure boot, or TPM settings don't match.
+    """
```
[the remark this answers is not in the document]
krcmarik: Per CLAUDE.md docstring rules: 'Returns: include only when returning a meaningful non-None value (omit for __init__ and other void functions).' This function returns None (void), so the Returns section is correctly omitted.

libs/providers/openshift.py
```diff
@@ -192,8 +192,27 @@ def vm_dict(
...
+
+        bootloader = firmware_spec.get("bootloader") if firmware_spec else None
+        efi_config = bootloader.get("efi") if bootloader else None
+
+        if efi_config:
+            firmware_info["boot_firmware"] = "efi"
+            firmware_info["secure_boot"] = efi_config["secureBoot"]
+        else:
+            firmware_info["boot_firmware"] = "bios"
+            firmware_info["secure_boot"] = False
```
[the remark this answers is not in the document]
krcmarik: Fixed. secureBoot access now uses .get() with None validation and raises ValueError with VM name context if absent. Also added type annotations to bootloader, efi_config, and secure_boot variables.

libs/providers/vmware.py
```diff
@@ -729,6 +729,16 @@ def vm_dict(self, **kwargs: Any) -> dict[str, Any]:
         # Guest OS
         result_vm_info["win_os"] = "win" in vm_config.guestId
 
+        firmware_info: dict[str, Any] = {}
+        firmware_info["boot_firmware"] = vm_config.firmware
+        boot_options = vm_config.bootOptions
+        firmware_info["secure_boot"] = boot_options.efiSecureBootEnabled if boot_options else False
```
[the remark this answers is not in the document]
krcmarik: In pyVmomi, bootOptions is either a vim.vm.BootOptions object or None. It cannot be an empty dict or other falsy value — truthiness check is semantically correct here. This is a pyVmomi attribute, not a dict .get() call.

libs/providers/openshift.py
```diff
@@ -192,8 +192,27 @@ def vm_dict(
...
-        firmware_spec: dict[str, Any] | None = cnv_vm.instance.spec.template.spec.domain.get("firmware")
+        domain: dict[str, Any] = cnv_vm.instance.spec.template.spec.domain
+
+        firmware_spec: dict[str, Any] | None = domain.get("firmware")
         result_vm_info["serial"] = firmware_spec.get("serial") if firmware_spec else None
+        firmware_info: dict[str, Any] = {}
+
+        devices: dict[str, Any] = domain.get("devices", {})
+        tpm_config: dict[str, Any] | None = devices.get("tpm")
+        firmware_info["tpm_present"] = bool(tpm_config.get("persistent")) if tpm_config else False
```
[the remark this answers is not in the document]
krcmarik: Fixed. TPM detection now uses tpm_config is not None to check for device presence, instead of checking the persistent flag.

libs/providers/vmware.py
```diff
@@ -729,6 +729,16 @@ def vm_dict(self, **kwargs: Any) -> dict[str, Any]:
...
+
```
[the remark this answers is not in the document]
krcmarik: pyVmomi's efiSecureBootEnabled already returns a proper Python bool. No normalization needed — the comparison with the OpenShift side (which also produces a bool) is type-safe.

libs/providers/openshift.py
```diff
@@ -192,8 +192,27 @@ def vm_dict(
...
+
```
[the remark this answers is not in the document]
krcmarik: Fixed. TPM detection now uses tpm_config is not None to check for device presence, as suggested.

libs/providers/openshift.py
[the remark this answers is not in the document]
krcmarik: Fixed. secureBoot access now uses .get() with None validation and raises ValueError if absent. We chose explicit validation over a silent default to fail fast per project guidelines.

utilities/post_migration.py
```diff
@@ -1548,6 +1587,11 @@ def check_vms(
             except Exception as exp:
                 res[vm_name].append(f"check_serial_preservation - {str(exp)}")
 
+            try:
+                check_firmware_and_tpm(source_vm=source_vm, destination_vm=destination_vm)
+            except Exception as exp:
+                res[vm_name].append(f"check_firmware_and_tpm - {str(exp)}")
```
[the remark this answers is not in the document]
krcmarik: This follows the established pattern in check_vms() where all validation checks use except Exception to collect all failures rather than stopping at the first one. Every other check in this function uses the same pattern. Changing just this one would be inconsistent.

libs/providers/openshift.py
```diff
@@ -192,8 +192,30 @@ def vm_dict(
...
+        bootloader: dict[str, Any] | None = firmware_spec.get("bootloader") if firmware_spec else None
+        efi_config: dict[str, Any] | None = bootloader.get("efi") if bootloader else None
+
+        if efi_config:
+            firmware_info["boot_firmware"] = "efi"
+            secure_boot: bool | None = efi_config.get("secureBoot")
+            if secure_boot is None:
+                raise ValueError(f"secureBoot not found in EFI config for VM '{cnv_vm_name}'")
+            firmware_info["secure_boot"] = secure_boot
```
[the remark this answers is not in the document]
krcmarik: The truthiness check is intentional here. KubeVirt can return empty/default EFI structures (efi: {}) for VMs that aren't actually EFI-booted. Changing to 'is not None' would misclassify those as EFI and then crash on the missing secureBoot key. We already hit this exact issue with the TPM detection — changing from a value check to 'is not None' caused false TPM mismatches in real migration tests because KubeVirt adds empty tpm: {} to non-TPM VMs. The truthiness check correctly treats empty objects as 'not configured'.

libs/providers/openshift.py
```diff
@@ -192,8 +192,30 @@ def vm_dict(
...
-        firmware_spec: dict[str, Any] | None = cnv_vm.instance.spec.template.spec.domain.get("firmware")
+        domain: dict[str, Any] = cnv_vm.instance.spec.template.spec.domain
+
+        firmware_spec: dict[str, Any] | None = domain.get("firmware")
         result_vm_info["serial"] = firmware_spec.get("serial") if firmware_spec else None
+        firmware_info: dict[str, Any] = {}
+
+        devices: dict[str, Any] = domain.get("devices", {})
+        tpm_config: dict[str, Any] | None = devices.get("tpm")
+        firmware_info["tpm_present"] = bool(tpm_config.get("persistent")) if tpm_config else False
```
myakove, line 206: **TPM detection semantic mismatch with VMware side**

`bool(tpm_config.get("persistent"))` checks if TPM is *persistent*, not if it *exists*. The VMware side (`libs/providers/vmware.py:734`) uses `isinstance(device, vim.vm.device.VirtualTPM)` — which checks device *existence*.

A TPM device with `persistent: false` would report `tpm_present=False` here but `True` on VMware → false assertion failure in `check_firmware_and_tpm`.

**Fix:** Use `tpm_config is not None` to match VMware's "device exists" semantics:
```python
firmware_info["tpm_present"] = tpm_config is not None
```
If persistence matters, track it as a separate field.

libs/providers/openshift.py
```diff
@@ -192,8 +192,30 @@ def vm_dict(
...
+
+        bootloader: dict[str, Any] | None = firmware_spec.get("bootloader") if firmware_spec else None
+        efi_config: dict[str, Any] | None = bootloader.get("efi") if bootloader else None
+
+        if efi_config:
+            firmware_info["boot_firmware"] = "efi"
+            secure_boot: bool | None = efi_config.get("secureBoot")
+            if secure_boot is None:
```
myakove, line 214: **`ValueError` in `vm_dict()` crashes entire VM check loop — bypasses error-collection pattern**

This `raise ValueError` executes inside `vm_dict()` (data collection phase), not inside `check_firmware_and_tpm()`. The `check_vms()` error-collection pattern (`except Exception`) only wraps individual check functions — it never reaches this point because `vm_dict()` is called earlier during VM data building.

If `secureBoot` is absent from an EFI config, this crashes data collection for ALL VMs, skipping every check (not just firmware).

**Fix:** Either:
1. Default to `False` with a warning in `vm_dict()`, then validate in `check_firmware_and_tpm()` where the error-collection pattern can catch it
2. Or move this validation into `check_firmware_and_tpm()` and set a sentinel value here

libs/providers/vmware.py
```diff
@@ -729,6 +729,16 @@ def vm_dict(self, **kwargs: Any) -> dict[str, Any]:
         # Guest OS
         result_vm_info["win_os"] = "win" in vm_config.guestId
 
+        firmware_info: dict[str, Any] = {}
```
myakove, line 734: **`bootOptions` silently defaults `secure_boot` to `False` when `None`**

If `vm_config.bootOptions` is `None` (e.g., VM in certain states), `secure_boot` silently becomes `False`. Per project guidelines "Fail Fast — Validate Content Not Just Existence": if a VM is EFI-firmware but has no `bootOptions`, that's a data integrity issue that should raise, not silently default.

**Fix:**
```python
boot_options = vm_config.bootOptions
if vm_config.firmware == "efi" and boot_options is None:
    raise ValueError(f"bootOptions is None for EFI VM '{vm_name}' — expected boot configuration")
firmware_info["secure_boot"] = boot_options.efiSecureBootEnabled if boot_options else False
```

utilities/post_migration.py
```diff
@@ -1170,6 +1170,45 @@ def check_serial_preservation(
...
+        source_vm: Source VM info dictionary with firmware data.
+        destination_vm: Destination VM info dictionary with firmware data.
+
+    Raises:
+        AssertionError: If firmware type, secure boot, or TPM settings don't match.
+    """
+    source_firmware: dict[str, Any] = source_vm["firmware"]
+    dest_firmware: dict[str, Any] = destination_vm["firmware"]
+
+    source_boot = source_firmware["boot_firmware"]
```
myakove: **Repetitive comparison pattern — DRY candidate**

Three identical compare-and-assert blocks for `boot_firmware`, `secure_boot`, `tpm_present` (18 lines). Could reduce to ~5 lines:

```python
checks = [("boot_firmware", "Boot firmware"), ("secure_boot", "Secure boot"), ("tpm_present", "TPM")]
for key, label in checks:
    assert source_firmware[key] == dest_firmware[key], (
        f"{label} mismatch for VM '{destination_vm['name']}': "
        f"source={source_firmware[key]}, destination={dest_firmware[key]}"
    )
```

Also makes adding new firmware checks (e.g., SMM) trivial.

libs/providers/vmware.py
```diff
@@ -729,6 +729,16 @@ def vm_dict(self, **kwargs: Any) -> dict[str, Any]:
         # Guest OS
         result_vm_info["win_os"] = "win" in vm_config.guestId
```
myakove, line 732: **Case-sensitive firmware string comparison without normalization**

`vm_config.firmware` returns `"bios"` or `"efi"` from the vSphere API. The OCP side (`openshift.py:209`) hardcodes `"efi"`/`"bios"`. The comparison in `check_firmware_and_tpm` uses `==` (case-sensitive).

If vSphere ever returns a different casing (e.g., `"EFI"`), the assertion would fail with a misleading mismatch error.

**Fix:** Normalize on at least one side:
```python
firmware_info["boot_firmware"] = vm_config.firmware.lower()
```
Or define shared constants (`FIRMWARE_EFI = "efi"`, `FIRMWARE_BIOS = "bios"`) used by both providers.

myakove (commented):
## Code Review

### :bulb: Suggestions (5)

| File | Line | Issue |
|------|------|-------|
| `libs/providers/openshift.py` | 203 | **TPM detection semantic mismatch with VMware side** |
| `libs/providers/openshift.py` | 211 | **`ValueError` in `vm_dict()` crashes entire VM check loop — bypasses error-coll |
| `libs/providers/vmware.py` | 732 | **`bootOptions` silently defaults `secure_boot` to `False` when `None`** |
| `utilities/post_migration.py` | 1186 | **Repetitive comparison pattern — DRY candidate** |
| `libs/providers/vmware.py` | 731 | **Case-sensitive firmware string comparison without normalization** |

---
*Review generated by pi*

libs/providers/openshift.py
krcmarik: We actually tried this exact change (`tpm_config is not None`) in an earlier commit and had to revert it — it caused a regression in real migration tests. KubeVirt adds empty `tpm: {}` to migrated VMs even when the source VM had no TPM device, so `tpm_config is not None` evaluates to `True` for VMs without TPM, producing false TPM mismatches.

The current `bool(tpm_config.get("persistent"))` correctly handles this: an empty `tpm: {}` has no `persistent` key, so `.get("persistent")` returns `None`, `bool(None)` is `False`, and the check correctly reports no TPM. Only VMs with an actual TPM device have `persistent: true` in the config.

libs/providers/openshift.py
krcmarik: `vm_dict()` is called per individual VM, not as a batch — if one VM's EFI config is missing `secureBoot`, only that VM's data collection fails, not all VMs.

The Fail Fast principle (project guideline) says: if an EFI VM has no `secureBoot` in its config, that's a real data integrity issue that should fail immediately with a clear error, not silently default to `False`. Defaulting would mask a broken VM config and produce misleading test results.

The `ValueError` with a descriptive message (`secureBoot not found in EFI config for VM '{name}'`) provides the exact context needed to debug the issue.

libs/providers/vmware.py
krcmarik: Fixed — added EFI-specific validation: `ValueError` is raised when an EFI VM has `bootOptions=None`. BIOS VMs continue to default `secure_boot` to `False`.

utilities/post_migration.py
krcmarik: Fixed — replaced the 3 repetitive comparison blocks with a loop over `firmware_checks`.

libs/providers/vmware.py
krcmarik: Fixed — added `.lower()` normalization to `vm_config.firmware`.

libs/providers/vmware.py
```diff
@@ -729,6 +729,19 @@ def vm_dict(self, **kwargs: Any) -> dict[str, Any]:
...
 
+        firmware_info: dict[str, Any] = {}
+        boot_firmware = vm_config.firmware.lower()
+        firmware_info["boot_firmware"] = boot_firmware
+        boot_options = vm_config.bootOptions
+        if boot_firmware == "efi" and boot_options is None:
+            raise ValueError(f"bootOptions is None for EFI VM '{_vm.name}' — expected boot configuration")
```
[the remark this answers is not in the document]
krcmarik: This is the same concern as the review comment on `openshift.py:211`, applied to the VMware side. The `ValueError` for EFI VMs with `bootOptions=None` is intentional Fail Fast behavior — `bootOptions` is always present for VMware VMs, so a `None` value indicates a genuine data integrity issue that should fail immediately rather than be silently swallowed. `vm_dict()` is called per individual VM, so one VM's failure doesn't affect others.

libs/providers/vmware.py
```diff
@@ -729,6 +729,19 @@ def vm_dict(self, **kwargs: Any) -> dict[str, Any]:
...
+        firmware_info["secure_boot"] = boot_options.efiSecureBootEnabled if boot_options else False
+
```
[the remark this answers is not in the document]
krcmarik: `vm_config.firmware` is a typed pyVmomi attribute that always returns `"bios"` or `"efi"` — it is never `None` or non-string. Adding `isinstance(firmware_value, str)` checks for VMware SDK attributes is unnecessary defensive coding. Similarly, `efiSecureBootEnabled` is always a boolean when `bootOptions` exists. We already validate `bootOptions is not None` for EFI VMs, which is the actual edge case worth guarding.

redhat-qe-bot1:
**Clean rebase detected** — no code changes compared to previous head (`90fda52`).
The following labels were preserved: `commented-krcmarik`, `commented-coderabbitai[bot]`, `lgtm-coderabbitai[bot]`, `commented-qodo-code-review[bot]`.

libs/providers/openshift.py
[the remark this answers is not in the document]
krcmarik: Skipped: Duplicate of review comment on line 211 — same secureBoot ValueError concern, already explained and rejected

utilities/post_migration.py
```diff
@@ -1635,6 +1662,11 @@ def check_vms(
             except Exception as exp:
                 res[vm_name].append(f"check_serial_preservation - {str(exp)}")
 
+            try:
+                check_firmware_and_tpm(source_vm=source_vm, destination_vm=destination_vm)
+            except Exception as exp:
```
myakove, line 1856: **`except Exception` wrapping `check_firmware_and_tpm`**

Per "No `except Exception` (MUST)" — `check_firmware_and_tpm` only raises `AssertionError` (from assert) or `KeyError` (if `"firmware"` key missing from VM dict). While this follows the pre-existing `check_vms()` pattern, new code shouldn't add new broad exceptions.

**Fix:** `except (AssertionError, KeyError) as exp:`

libs/providers/vmware.py
```diff
@@ -729,6 +729,19 @@ def vm_dict(self, **kwargs: Any) -> dict[str, Any]:
         # Guest OS
         result_vm_info["win_os"] = "win" in vm_config.guestId
 
+        firmware_info: dict[str, Any] = {}
+        boot_firmware = vm_config.firmware.lower()
```
myakove: **`vm_config.firmware` accessed without validation on external provider data**

If `firmware` is `None` (unusual VM state), `.lower()` raises `AttributeError`. Per "Fail Fast — Validate Content Not Just Existence" for external data.

Also: `boot_firmware` and `boot_options` lack type annotations — the parallel OpenShift code thoroughly annotates every local variable.

**Fix:**
```python
boot_firmware_raw: str | None = vm_config.firmware
if not boot_firmware_raw:
    raise ValueError(f"firmware attribute missing for VM '{_vm.name}'")
boot_firmware: str = boot_firmware_raw.lower()
```

utilities/post_migration.py
```diff
@@ -1257,6 +1257,33 @@ def check_serial_preservation(
...
+        source_vm: Source VM info dictionary with firmware data.
+        destination_vm: Destination VM info dictionary with firmware data.
+
+    Raises:
+        AssertionError: If firmware type, secure boot, or TPM settings don't match.
+    """
+    source_firmware: dict[str, Any] = source_vm["firmware"]
+    dest_firmware: dict[str, Any] = destination_vm["firmware"]
+
+    firmware_checks = [("boot_firmware", "Boot firmware"), ("secure_boot", "Secure boot"), ("tpm_present", "TPM")]
```
myakove: **Missing type annotation on `firmware_checks`**

Per project style that annotates data structures:
```python
firmware_checks: list[tuple[str, str]] = [("boot_firmware", "Boot firmware"), ...]
```

utilities/post_migration.py
```diff
@@ -1257,6 +1257,33 @@ def check_serial_preservation(
         LOGGER.info(f"Serial preserved correctly (OCP {ocp_version} < 4.20, plain UUID): {dest_serial}")
 
 
+def check_firmware_and_tpm(source_vm: dict[str, Any], destination_vm: dict[str, Any]) -> None:
```
myakove: **[SUGGESTION]** Function name `check_firmware_and_tpm` uses "and".

Per "Single responsibility — if you need 'and' to describe it in the docstring, split it". Since firmware type, secure boot, and TPM are all boot configuration metadata, consider renaming to `check_boot_configuration` — avoids the "and" while keeping them grouped under a single logical concept.

myakove (commented):
## Code Review

Found **4** issue(s) in this PR:

### :bulb: Suggestions (4)

| File | Line | Issue |
|------|------|-------|
| `utilities/post_migration.py` | 1667 | **`except Exception` wrapping `check_firmware_and_tpm`** |
| `libs/providers/vmware.py` | 733 | **`vm_config.firmware` accessed without validation on external provider data** |
| `utilities/post_migration.py` | 1273 | **Missing type annotation on `firmware_checks`** |
| `utilities/post_migration.py` | 1260 | **[SUGGESTION]** Function name `check_firmware_and_tpm` uses "and". |

---
*Review generated by pi*

utilities/post_migration.py
krcmarik: Fixed. Narrowed to `except (AssertionError, KeyError) as exp:` — these are the only two exception types `check_boot_configuration` can raise.

libs/providers/vmware.py
krcmarik: Fixed. Added fail-fast validation with `ValueError` before `.lower()` call, and added type annotations for `boot_firmware_raw: str | None`, `boot_firmware: str`, and `boot_options: vim.vm.BootOptions | None`.

utilities/post_migration.py
krcmarik: Fixed. Added `list[tuple[str, str]]` type annotation.

utilities/post_migration.py
krcmarik: Fixed. Renamed to `check_boot_configuration` — function definition, call site, error message, and docstring all updated.

utilities/post_migration.py
```diff
@@ -1635,6 +1666,11 @@ def check_vms(
             except Exception as exp:
                 res[vm_name].append(f"check_serial_preservation - {str(exp)}")
 
+            try:
+                check_boot_configuration(source_vm=source_vm, destination_vm=destination_vm)
+            except (AssertionError, KeyError) as exp:
```
myakove: **Exception tuple missing `ValueError` — can crash `check_vms()` loop**

The firmware extractors in `openshift.py:211` and `vmware.py:734` raise `ValueError` for invalid firmware config (e.g., missing secureBoot, missing firmware attribute). If `ValueError` propagates uncaught here, it crashes the entire `check_vms()` loop and prevents remaining VMs from being checked.

**Fix:** Add `ValueError` to the tuple:
```python
except (AssertionError, KeyError, ValueError) as exp:
```

libs/providers/vmware.py
```diff
@@ -729,6 +729,22 @@ def vm_dict(self, **kwargs: Any) -> dict[str, Any]:
...
+        firmware_info: dict[str, Any] = {}
+        boot_firmware_raw: str | None = vm_config.firmware
+        if not boot_firmware_raw:
+            raise ValueError(f"firmware attribute is missing or empty for VM '{_vm.name}'")
+        boot_firmware: str = boot_firmware_raw.lower()
+        firmware_info["boot_firmware"] = boot_firmware
+        boot_options: vim.vm.BootOptions | None = vm_config.bootOptions
+        if boot_firmware == "efi" and boot_options is None:
+            raise ValueError(f"bootOptions is None for EFI VM '{_vm.name}' — expected boot configuration")
+        firmware_info["secure_boot"] = boot_options.efiSecureBootEnabled if boot_options else False
```
myakove: **[SUGGESTION]** BIOS VMs read `efiSecureBootEnabled` unnecessarily.

For BIOS VMs, `boot_options.efiSecureBootEnabled` is an EFI-specific field with no meaningful value. The OCP side explicitly sets `secure_boot=False` for non-EFI. This works because the value is always `False` for BIOS, but it's semantically fragile.

**Fix:** Guard the EFI-specific read:
```python
if boot_firmware == "efi":
    firmware_info["secure_boot"] = boot_options.efiSecureBootEnabled if boot_options else False
else:
    firmware_info["secure_boot"] = False
```

Also: PR description still says `check_firmware_and_tpm()` but the function was renamed to `check_boot_configuration()`, and says "all migration tests" but it's only vSphere (Group 3). Please update the description.

myakove (commented):
## Code Review

### :bulb: Suggestions (2)

| File | Line | Issue |
|------|------|-------|
| `utilities/post_migration.py` | 1671 | **Exception tuple missing `ValueError` — can crash `check_vms()` loop** |
| `libs/providers/vmware.py` | 741 | **[SUGGESTION]** BIOS VMs read `efiSecureBootEnabled` unnecessarily. |

---
*Review generated by pi*

utilities/post_migration.py
krcmarik: Good catch — reverted to except Exception to match the established pattern in check_vms(). The narrowing to (AssertionError, KeyError) in the previous commit was a mistake — check_vms() is an aggregation function that collects errors from all checks and reports them together. Narrowing the except clause risks crashing the entire loop on an unexpected exception type, preventing remaining VMs from being checked. All other 16 check blocks use except Exception for this reason.

libs/providers/vmware.py
krcmarik: Guarded efiSecureBootEnabled read with firmware type check — BIOS VMs now get secure_boot=False directly. Also simplified the EFI branch by removing the dead boot_options ternary (guaranteed non-None by the ValueError guard above). Will update PR description separately.

krcmarik:
@coderabbitai[bot]

The following review comments were reviewed and a decision was made:

### `utilities/post_migration.py:1267` (outside diff) — LOW: `Raises` section is incomplete for `check_boot_configuration`
> Addressed: Added KeyError to the Raises section — documents the KeyError from source_vm['firmware'] and dest_firmware[key] bracket access.
### `utilities/post_migration.py:1267` (outside diff) — LOW: `Raises` section is incomplete for `check_boot_configuration`
> Addressed: Duplicate of the above — addressed together.

utilities/post_migration.py
```diff
@@ -1635,6 +1667,11 @@ def check_vms(
             except Exception as exp:
                 res[vm_name].append(f"check_serial_preservation - {str(exp)}")
 
+            try:
+                check_boot_configuration(source_vm=source_vm, destination_vm=destination_vm)
+            except Exception as exp:
+                res[vm_name].append(f"check_boot_configuration - {str(exp)}")
```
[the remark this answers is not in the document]
krcmarik: The firmware dict is our controlled data, populated by our own provider extractors. Per project guidelines, controlled config uses bracket access and fails fast with KeyError — adding .get() validation is the anti-pattern we explicitly avoid. KeyError('firmware') tells you exactly which key is missing, and the 'check_boot_configuration' prefix in the error identifies which check failed. The error messages are not opaque — they follow the same pattern as all other checks in check_vms().

krcmarik:
@coderabbitai[bot]

The following review comments were reviewed and a decision was made:

### `utilities/post_migration.py:1274` (outside diff) — HIGH: `tpm_present` is comparing different semantics across providers.
> Skipped: The semantics are intentionally correct. KubeVirt adds an empty tpm: {} to ALL migrated VMs regardless of whether the source had TPM — this was discovered and fixed in a previous iteration. Checking tpm_config is not None was tried and failed for this exact reason. The persistent flag is KubeVirt/Forklift's way of indicating actual TPM device presence in the declarative spec model, not a separate persistence setting. Both providers correctly answer 'does this VM have a TPM device?' — the difference is purely representational (object enumeration vs declarative spec).
### `utilities/post_migration.py:1670` (duplicate) — MEDIUM: Don't broaden this catch to `Exception`.
> Skipped: check_vms() is an error aggregation function — all 17 check blocks use except Exception to collect errors and report them together via pytest.fail(). Narrowing this one block risks crashing the entire loop on unexpected exceptions, preventing remaining VMs from being checked. This was previously narrowed to (AssertionError, KeyError) based on a review comment and has been intentionally reverted to match the established pattern.

libs/providers/vmware.py
```diff
@@ -729,6 +729,26 @@ def vm_dict(self, **kwargs: Any) -> dict[str, Any]:
...
+        firmware_info: dict[str, Any] = {}
+        boot_firmware_raw: str | None = vm_config.firmware
+        if not boot_firmware_raw:
+            raise ValueError(f"firmware attribute is missing or empty for VM '{_vm.name}'")
+        boot_firmware: str = boot_firmware_raw.lower()
+        firmware_info["boot_firmware"] = boot_firmware
+        boot_options: vim.vm.BootOptions | None = vm_config.bootOptions
+        if boot_firmware == "efi" and boot_options is None:
+            raise ValueError(f"bootOptions is None for EFI VM '{_vm.name}' — expected boot configuration")
+        if boot_firmware == "efi":
```
myakove, line 743: **Duplicate conditional + unnecessary `assert` for type narrowing**

`if boot_firmware == "efi"` is checked twice in separate blocks (line 739 and 741), bridged by `assert boot_options is not None` (line 742) for mypy. The `assert` is stripped with `-O` and the condition is checked twice.

**Fix:** Combine into one block:
```python
if boot_firmware == "efi":
    if boot_options is None:
        raise ValueError(f"bootOptions is None for EFI VM '{_vm.name}' — expected boot configuration")
    firmware_info["secure_boot"] = boot_options.efiSecureBootEnabled
else:
    firmware_info["secure_boot"] = False
```
This removes the duplicate check, eliminates the `assert`, and is cleaner for both mypy and runtime.

libs/providers/vmware.py
```diff
@@ -729,6 +729,26 @@ def vm_dict(self, **kwargs: Any) -> dict[str, Any]:
         # Guest OS
         result_vm_info["win_os"] = "win" in vm_config.guestId
 
+        firmware_info: dict[str, Any] = {}
+        boot_firmware_raw: str | None = vm_config.firmware
```
myakove, line 735: **[SUGGESTION]** PR description still outdated (asked in cycles 2 and 3).

Description says `check_firmware_and_tpm()` but function is `check_boot_configuration()`. Also says "all migration tests" but it only runs for vSphere (Group 3).

Please update the PR description to match the current implementation.

myakove (commented):
## Code Review

### :bulb: Suggestions (2)

| File | Line | Issue |
|------|------|-------|
| `libs/providers/vmware.py` | 741 | **Duplicate conditional + unnecessary `assert` for type narrowing** |
| `libs/providers/vmware.py` | 733 | **[SUGGESTION]** PR description still outdated (asked in cycles 2 and 3). |

---
*Review generated by pi*

libs/providers/vmware.py
krcmarik: Done. Combined the duplicate `if boot_firmware == "efi"` into a single block with the `ValueError` guard nested inside — removes the assert and the redundant condition check.

libs/providers/vmware.py
krcmarik: Updated — fixed the function name to `check_boot_configuration()` and clarified scope to vSphere migration tests.

krcmarik:
@coderabbitai review

krcmarik:
@coderabbitai review

krcmarik:
@coderabbitai full review

myakove (commented):
## Code Review

### :bulb: Suggestions (5)

| File | Line | Issue |
|------|------|-------|
| `libs/providers/openshift.py` | 203 | **TPM detection uses `persistent` value instead of device existence — semantic m |
| `libs/providers/openshift.py` | 205 | **Silent BIOS fallback when `firmware_spec` exists but has no `bootloader` key** |
| `libs/providers/openshift.py` | 199 | **[SUGGESTION]** `vm_dict` is 218 lines (VMware's is 116) — both exceed the 50-l |
| `libs/providers/vmware.py` | 746 | **[SUGGESTION]** Duplicate iteration over `vm_config.hardware.device` for TPM ch |
| `libs/providers/vmware.py` | 732 | **[SUGGESTION]** `vm_dict` is 116 lines — consider extracting firmware extractio |

---
*Review generated by pi*

libs/providers/openshift.py
myakove, line 206: **TPM detection uses `persistent` value instead of device existence — semantic mismatch with VMware**

VMware checks TPM *existence* (`isinstance(device, vim.vm.device.VirtualTPM)`), but this checks the `persistent` sub-key's *value*. KubeVirt adds `tpm: {}` (without `persistent`) to migrated VMs even when the source had a TPM device. Result:
- VMware source: `tpm_present = True` (VirtualTPM device exists)
- OCP destination: `tpm_present = False` (`persistent` key missing → `.get()` returns None → `bool(None)` = False)

The author previously explained that `tpm_config is not None` doesn't work because KubeVirt also adds empty `tpm: {}` for VMs *without* TPM. So `persistent` is the reliable discriminator. If that's the case, add a comment explaining *why* `persistent` is used instead of mere presence:

```suggestion
        firmware_info["tpm_present"] = bool(tpm_config.get("persistent")) if tpm_config else False  # KubeVirt adds empty tpm: {} even for VMs without TPM; only persistent=true indicates actual TPM
```

libs/providers/openshift.py
```diff
@@ -192,8 +192,30 @@ def vm_dict(
...
+
+        firmware_spec: dict[str, Any] | None = domain.get("firmware")
         result_vm_info["serial"] = firmware_spec.get("serial") if firmware_spec else None
+        firmware_info: dict[str, Any] = {}
+
+        devices: dict[str, Any] = domain.get("devices", {})
+        tpm_config: dict[str, Any] | None = devices.get("tpm")
+        firmware_info["tpm_present"] = bool(tpm_config.get("persistent")) if tpm_config else False
+
+        bootloader: dict[str, Any] | None = firmware_spec.get("bootloader") if firmware_spec else None
```
myakove, line 208: **Silent BIOS fallback when `firmware_spec` exists but has no `bootloader` key**

If a KubeVirt EFI VM has a `firmware` spec (e.g., containing only `serial`) but no `bootloader` key — due to a Forklift bug, API change, or manual edit — `bootloader` is `None`, `efi_config` is `None`, and the code falls into the `else` branch reporting `boot_firmware = "bios"`. This masks the real problem (missing bootloader config) behind a misleading firmware-type mismatch assertion.

**Suggestion:** Consider adding a warning when `firmware_spec` exists but has no bootloader.

libs/providers/openshift.py
```diff
@@ -192,8 +192,30 @@ def vm_dict(
         )
 
         # Serial number (from firmware) - for serial preservation verification
-        firmware_spec: dict[str, Any] | None = cnv_vm.instance.spec.template.spec.domain.get("firmware")
+        domain: dict[str, Any] = cnv_vm.instance.spec.template.spec.domain
+
+        firmware_spec: dict[str, Any] | None = domain.get("firmware")
         result_vm_info["serial"] = firmware_spec.get("serial") if firmware_spec else None
+        firmware_info: dict[str, Any] = {}
```
myakove, line 201: **[SUGGESTION]** `vm_dict` is 218 lines (VMware's is 116) — both exceed the 50-line SHOULD guideline.

Pre-existing tech debt, but the firmware block (lines 199-218) is a strong extraction candidate into `_extract_firmware_info(domain: dict) -> dict[str, Any]`.

Also: missing section comment — other sections use `# Power state`, `# Serial number`, `# VM labels`. Add before this line:
```python
# Firmware configuration — boot type, secure boot, TPM
```

libs/providers/vmware.py
```diff
@@ -729,6 +729,25 @@ def vm_dict(self, **kwargs: Any) -> dict[str, Any]:
...
+        firmware_info["boot_firmware"] = boot_firmware
+        boot_options: vim.vm.BootOptions | None = vm_config.bootOptions
+        if boot_firmware == "efi":
+            if boot_options is None:
+                raise ValueError(f"bootOptions is None for EFI VM '{_vm.name}' — expected boot configuration")
+            firmware_info["secure_boot"] = boot_options.efiSecureBootEnabled
+        else:
+            firmware_info["secure_boot"] = False
+
+        firmware_info["tpm_present"] = any(
```
myakove, line 750: **[SUGGESTION]** Duplicate iteration over `vm_config.hardware.device` for TPM check.

The device list is already iterated earlier (~line 677) for NICs and disks. The TPM `isinstance` check could be folded into that existing loop to avoid a second full pass. Not a bug — minor performance/consistency improvement.

libs/providers/vmware.py
```diff
@@ -729,6 +729,25 @@ def vm_dict(self, **kwargs: Any) -> dict[str, Any]:
         # Guest OS
         result_vm_info["win_os"] = "win" in vm_config.guestId
 
+        firmware_info: dict[str, Any] = {}
```
myakove, line 734: **[SUGGESTION]** `vm_dict` is 116 lines — consider extracting firmware extraction.

The firmware block (lines 732-749) could be a private helper `_extract_firmware_info(vm_config, vm_name) -> dict[str, Any]`. Same as the OpenShift side — makes the parallel structure explicit.

Also consider adding `"firmware": {}` to `VIRTUAL_MACHINE_TEMPLATE` in `base_provider.py` to make the contract explicit across all providers.

libs/providers/openshift.py
krcmarik: Added inline comment explaining why `persistent` is used instead of device existence: `# KubeVirt adds empty tpm: {} even for VMs without TPM; only persistent=true indicates actual TPM`

This documents the non-obvious design decision that caused a revert cycle in earlier review rounds (we tried `tpm_config is not None` but KubeVirt adds empty `tpm: {}` for all VMs regardless of TPM presence).

libs/providers/openshift.py
krcmarik: Skipped — theoretical edge case. Forklift always sets `bootloader` for EFI VMs. If `firmware_spec` exists with only `serial` and no `bootloader`, KubeVirt defaults to BIOS, which is the correct behavior (no bootloader config = BIOS). The scenario described (Forklift bug or API change removing bootloader from an EFI VM) would require a Forklift fix, not a warning in test code.

libs/providers/openshift.py
krcmarik: Added section comment `# Firmware configuration — boot type, secure boot, TPM` before the firmware block, matching the existing pattern (`# Power state`, `# Serial number`, `# VM labels`).

Regarding extraction into `_extract_firmware_info()`: agreed this is a valid refactoring candidate, but it's pre-existing tech debt affecting both providers. Better suited for a dedicated refactoring PR to keep this PR focused on the firmware verification feature.

libs/providers/vmware.py
krcmarik: Skipped — the device list for a typical VM has ~10-20 items, so a second pass is trivial. Folding TPM detection into the NIC/disk loop would mix unrelated concerns (networking/storage vs security) and reduce readability. The current structure groups by concept: NICs, disks, then TPM separately.

libs/providers/vmware.py
krcmarik: Added section comment `# Firmware configuration — boot type, secure boot, TPM` before the firmware block, matching the existing pattern (`# CPUs`, `# Memory`, `# Guest OS`).

Regarding extraction and adding `firmware: {}` to `VIRTUAL_MACHINE_TEMPLATE`: agreed both are valid improvements, but they're refactoring scope that touches base_provider.py (shared by all providers). Better suited for a dedicated refactoring PR.

myakove (commented):
## Code Review

### :bulb: Suggestions (6)

| File | Line | Issue |
|------|------|-------|
| `libs/providers/openshift.py` | 207 | ⚠️ **[PREV-UNRESOLVED] Silent BIOS fallback when `firmware_spec` exists but has  |
| `libs/providers/vmware.py` | 747 | 💡 **[PREV-UNRESOLVED] Duplicate iteration over `vm_config.hardware.device` for T |
| `utilities/post_migration.py` | 1280 | **Assert loop short-circuits: only first firmware mismatch reported per VM** |
| `libs/providers/openshift.py` | 213 | **`vm_dict` docstring `Raises` section is stale after new `ValueError`** |
| `libs/providers/vmware.py` | 737 | 💡 **Validate `boot_firmware` against known values at extraction time** |
| `libs/providers/openshift.py` | 198 | 💡 **Interleaved concerns reduce readability** |

---
*Review generated by pi*

libs/providers/openshift.py
```diff
@@ -192,8 +192,32 @@ def vm_dict(
...
         result_vm_info["serial"] = firmware_spec.get("serial") if firmware_spec else None
+        # Firmware configuration — boot type, secure boot, TPM
+        firmware_info: dict[str, Any] = {}
+
+        devices: dict[str, Any] = domain.get("devices", {})
+        tpm_config: dict[str, Any] | None = devices.get("tpm")
+        # KubeVirt adds empty tpm: {} even for VMs without TPM; only persistent=true indicates actual TPM
+        firmware_info["tpm_present"] = bool(tpm_config.get("persistent")) if tpm_config else False
+
+        bootloader: dict[str, Any] | None = firmware_spec.get("bootloader") if firmware_spec else None
```
myakove, line 208: ⚠️ **[PREV-UNRESOLVED] Silent BIOS fallback when `firmware_spec` exists but has no `bootloader` key**

If a KubeVirt EFI VM has a `firmware` spec (e.g., containing only `serial`) but no `bootloader` key — due to a Forklift bug, API change, or manual edit — `bootloader` is `None`, `efi_config` is `None`, and the code falls into the `else` branch reporting `boot_firmware = "bios"`. This masks the real problem (missing bootloader config) behind a misleading firmware-type mismatch assertion.

Your response ("theoretical edge case, Forklift always sets bootloader") is reasonable, but the thread was left unresolved. If this is intentional behavior, please resolve the thread with a brief note.

libs/providers/vmware.py
```diff
@@ -729,6 +729,26 @@ def vm_dict(self, **kwargs: Any) -> dict[str, Any]:
...
+        firmware_info["boot_firmware"] = boot_firmware
+        boot_options: vim.vm.BootOptions | None = vm_config.bootOptions
+        if boot_firmware == "efi":
+            if boot_options is None:
+                raise ValueError(f"bootOptions is None for EFI VM '{_vm.name}' — expected boot configuration")
+            firmware_info["secure_boot"] = boot_options.efiSecureBootEnabled
+        else:
+            firmware_info["secure_boot"] = False
+
+        firmware_info["tpm_present"] = any(
```
myakove, line 750: 💡 **[PREV-UNRESOLVED] Duplicate iteration over `vm_config.hardware.device` for TPM check**

The device list is already iterated earlier (~line 677) for NICs and disks. The TPM `isinstance` check could be folded into that existing loop to avoid a second full pass.

Your response ("trivial performance, mixing concerns reduces readability") is reasonable. If skipping intentionally, please resolve the thread.

utilities/post_migration.py
```diff
@@ -1257,6 +1257,38 @@ def check_serial_preservation(
...
+    source_firmware: dict[str, Any] = source_vm["firmware"]
+    dest_firmware: dict[str, Any] = destination_vm["firmware"]
+
+    firmware_checks: list[tuple[str, str]] = [
+        ("boot_firmware", "Boot firmware"),
+        ("secure_boot", "Secure boot"),
+        ("tpm_present", "TPM"),
+    ]
+    for key, label in firmware_checks:
+        assert source_firmware[key] == dest_firmware[key], (
```
myakove: **Assert loop short-circuits: only first firmware mismatch reported per VM**

The `assert` inside the for-loop stops on the first failure. If `boot_firmware` mismatches, you'll never learn that `secure_boot` or `TPM` also mismatch until you fix the first one and re-run. Since `check_vms()` catches the exception and moves to the next VM, you only get one firmware failure per VM per run.

Suggestion — collect all mismatches and raise once:
```suggestion
    mismatches: list[str] = []
    for key, label in firmware_checks:
        if source_firmware[key] != dest_firmware[key]:
            mismatches.append(
                f"{label}: source={source_firmware[key]}, destination={dest_firmware[key]}"
            )
    if mismatches:
        raise AssertionError(
            f"Firmware mismatch for VM '{destination_vm['name']}': {'; '.join(mismatches)}"
        )
```

libs/providers/openshift.py
```diff
@@ -192,8 +192,32 @@ def vm_dict(
...
+        efi_config: dict[str, Any] | None = bootloader.get("efi") if bootloader else None
+
+        if efi_config:
+            firmware_info["boot_firmware"] = "efi"
+            secure_boot: bool | None = efi_config.get("secureBoot")
+            if secure_boot is None:
```
myakove, line 214: **`vm_dict` docstring `Raises` section is stale after new `ValueError`**

The new `ValueError("secureBoot not found in EFI config for VM...")` here is not documented in the method's docstring `Raises` section (line ~127), which only mentions `ValueError` for "ocp_resource not set or VM fails to start".

Per "Docstring Format (MUST)": update the `Raises` section to document this additional failure mode:
```python
Raises:
    ValueError: If `ocp_resource` is not set, VM fails to start,
        or secureBoot is missing from an EFI VM's firmware config.
```

libs/providers/vmware.py
```diff
@@ -729,6 +729,26 @@ def vm_dict(self, **kwargs: Any) -> dict[str, Any]:
         # Guest OS
         result_vm_info["win_os"] = "win" in vm_config.guestId
 
+        # Firmware configuration — boot type, secure boot, TPM
+        firmware_info: dict[str, Any] = {}
+        boot_firmware_raw: str | None = vm_config.firmware
+        if not boot_firmware_raw:
+            raise ValueError(f"firmware attribute is missing or empty for VM '{_vm.name}'")
+        boot_firmware: str = boot_firmware_raw.lower()
```
myakove, line 738: 💡 **Validate `boot_firmware` against known values at extraction time**

If VMware returns an unexpected firmware string (e.g., `"uefi"` from a future API version), the mismatch would surface as a generic `AssertionError` in `check_boot_configuration` rather than a clear `ValueError` at the source.

Per "Fail Fast - Validate Content Not Just Existence (MUST)":
```suggestion
        boot_firmware: str = boot_firmware_raw.lower()
        if boot_firmware not in ("bios", "efi"):
            raise ValueError(f"Unexpected firmware type '{boot_firmware}' for VM '{_vm.name}'")
        firmware_info["boot_firmware"] = boot_firmware
```

libs/providers/openshift.py
```diff
@@ -192,8 +192,32 @@ def vm_dict(
         )
 
         # Serial number (from firmware) - for serial preservation verification
-        firmware_spec: dict[str, Any] | None = cnv_vm.instance.spec.template.spec.domain.get("firmware")
+        domain: dict[str, Any] = cnv_vm.instance.spec.template.spec.domain
+
+        firmware_spec: dict[str, Any] | None = domain.get("firmware")
         result_vm_info["serial"] = firmware_spec.get("serial") if firmware_spec else None
```
myakove, line 199: 💡 **Interleaved concerns reduce readability**

Serial extraction (line 198) and bootloader extraction (line 207) both read from `firmware_spec`, but they're separated by the TPM extraction block (lines 202-205) which reads from `devices`. Grouping `firmware_spec`-derived fields together and `devices`-derived fields together would improve readability and make the data flow clearer.

Suggestion: reorder so serial + bootloader/EFI extraction is contiguous, followed by TPM extraction from devices.

libs/providers/openshift.py
krcmarik: Confirmed — intentional skip. Resolving per follow-up request.

libs/providers/vmware.py
krcmarik: Confirmed — intentional skip. Resolving per follow-up request.

libs/providers/openshift.py
krcmarik: Resolving — see prior thread for full reasoning (theoretical edge case, Forklift always sets bootloader for EFI VMs).

libs/providers/vmware.py
krcmarik: Resolving — see prior thread for full reasoning (trivial perf, separation of concerns is more readable).

utilities/post_migration.py
krcmarik: Fixed — `check_boot_configuration` now collects all firmware mismatches (boot_firmware, secure_boot, TPM) into a list and raises a single `AssertionError` with all of them joined by semicolons.

libs/providers/openshift.py
krcmarik: Fixed — updated `Raises` section to include secureBoot `ValueError` for EFI VMs.

libs/providers/vmware.py
krcmarik: Fixed — added validation that `boot_firmware` is either `"bios"` or `"efi"`, raising `ValueError` with VM name for unexpected values.

libs/providers/openshift.py
krcmarik: Skipped — the current ordering follows a logical flow: serial (from firmware_spec), then firmware_info block (TPM from devices, bootloader from firmware_spec). Section comments already clarify the grouping. Reordering would be pure style churn with no functional benefit.

krcmarik:
@coderabbitai[bot]

The following review comments were reviewed and a decision was made:

### `libs/providers/openshift.py:199` (outside diff) — HIGH: Replace `dict[str, Any]` with a concrete firmware contract.
> Not addressed: Skipped — `firmware_info` gets assigned into `result_vm_info["firmware"]` which is `dict[str, Any]` (from `VIRTUAL_MACHINE_TEMPLATE`), so the narrower type is immediately lost at the boundary. A proper fix would require `TypedDict` for the whole VM template, which is a larger refactor out of scope for this PR.

libs/providers/openshift.py
krcmarik: Confirmed — resolving. The current ordering is intentional: serial (firmware_spec), then firmware_info block with section comments. See prior reply for full reasoning.

krcmarik:
@coderabbitai[bot]

The following review comments were reviewed and a decision was made:

### `libs/providers/openshift.py:199` (outside diff) — HIGH: Replace `dict[str, Any]` with a concrete firmware contract.
> Not addressed: Skipped — `firmware_info` gets assigned into `result_vm_info["firmware"]` which is `dict[str, Any]` (from `VIRTUAL_MACHINE_TEMPLATE`), so the narrower type is immediately lost at the boundary. A proper fix would require `TypedDict` for the whole VM template, which is out of scope for this PR.

redhat-qe-bot2:
**Clean rebase detected** — no code changes compared to previous head (`1e12cd3`).
The following labels were preserved: `verified`, `commented-krcmarik`, `lgtm-coderabbitai[bot]`.

utilities/post_migration.py
```diff
@@ -1257,6 +1257,39 @@ def check_serial_preservation(
         LOGGER.info(f"Serial preserved correctly (OCP {ocp_version} < 4.20, plain UUID): {dest_serial}")
 
 
+def check_boot_configuration(source_vm: dict[str, Any], destination_vm: dict[str, Any]) -> None:
```
myakove, line 1368: ⚠️ **`VIRTUAL_MACHINE_TEMPLATE` in `libs/base_provider.py` missing `firmware` key**

The template defines the contract for all VM info dicts (id, name, disks, cpu, etc.) but the new `firmware` key is only set dynamically inside VMware and OCP providers. Other providers (RHV, OVA, OpenStack) won't have it.

This function accesses `source_vm["firmware"]` directly (line 1273) — it would crash with `KeyError` if ever called for a non-vSphere provider. Currently safe (gated by the vSphere check in `check_vms()`), but the template contract should be explicit.

**Fix:** Add `"firmware": {}` to `VIRTUAL_MACHINE_TEMPLATE` in `libs/base_provider.py`.

---

libs/providers/openshift.py
```diff
@@ -192,8 +193,32 @@ def vm_dict(
...
+        firmware_info["tpm_present"] = bool(tpm_config.get("persistent")) if tpm_config else False
+
+        bootloader: dict[str, Any] | None = firmware_spec.get("bootloader") if firmware_spec else None
+        efi_config: dict[str, Any] | None = bootloader.get("efi") if bootloader else None
+
+        if efi_config:
+            firmware_info["boot_firmware"] = "efi"
+            secure_boot: bool | None = efi_config.get("secureBoot")
+            if secure_boot is None:
+                raise ValueError(f"secureBoot not found in EFI config for VM '{cnv_vm_name}'")
```
myakove, line 215: ⚠️ **Non-EFI bootloader silently classified as BIOS**

If `bootloader` is present but has no `efi` key (e.g., a future KubeVirt bootloader type), the code falls to the `else` branch and reports `"bios"` with `secure_boot=False`. The VMware provider correctly rejects unexpected firmware types with `ValueError`.

The OCP provider should mirror this defensive pattern — raise when `bootloader` exists but isn't EFI, rather than silently defaulting to BIOS:

```suggestion
        if efi_config:
            firmware_info["boot_firmware"] = "efi"
            secure_boot: bool | None = efi_config.get("secureBoot")
            if secure_boot is None:
                raise ValueError(f"secureBoot not found in EFI config for VM '{cnv_vm_name}'")
            firmware_info["secure_boot"] = secure_boot
        elif bootloader:
            raise ValueError(
                f"Unexpected bootloader config (not EFI) for VM '{cnv_vm_name}': {bootloader}"
            )
        else:
            firmware_info["boot_firmware"] = "bios"
            firmware_info["secure_boot"] = False
```

---

libs/providers/vmware.py
```diff
@@ -729,6 +729,28 @@ def vm_dict(self, **kwargs: Any) -> dict[str, Any]:
         # Guest OS
         result_vm_info["win_os"] = "win" in vm_config.guestId
 
+        # Firmware configuration — boot type, secure boot, TPM
```
myakove, line 733: 💡 **[SUGGESTION] Firmware extraction inlined in already oversized `vm_dict`**

Both `vm_dict` methods are well over the 50-line guideline (VMware ~119 lines, OCP ~220 lines). The firmware extraction blocks (~22-24 lines each) are self-contained sub-tasks with clear input/output boundaries — ideal candidates for extraction.

Suggestion: Extract into `_extract_firmware_info(vm_config, vm_name)` (VMware) and `_extract_firmware_info(domain, cnv_vm_name)` (OCP). Improves readability and testability.

---

myakove (commented):
## Code Review

### :bulb: Suggestions (3)

| File | Line | Issue |
|------|------|-------|
| `utilities/post_migration.py` | 1260 | ⚠️ **`VIRTUAL_MACHINE_TEMPLATE` in `libs/base_provider.py` missing `firmware` ke |
| `libs/providers/openshift.py` | 215 | ⚠️ **Non-EFI bootloader silently classified as BIOS** |
| `libs/providers/vmware.py` | 732 | 💡 **[SUGGESTION] Firmware extraction inlined in already oversized `vm_dict`** |

---
*Review generated by pi*

---

utilities/post_migration.py
krcmarik: Done. Added `"firmware": {}` to `VIRTUAL_MACHINE_TEMPLATE` in `libs/base_provider.py`.

Note: this makes the contract explicit but doesn't fully prevent `KeyError` on its own — `check_boot_configuration` accesses nested keys like `source_firmware["boot_firmware"]` which would still fail on the empty dict. The real guard is the vSphere provider-type gate in `check_vms()` that prevents the function from being called for non-vSphere providers. The template addition prepares for future provider firmware support.

libs/providers/openshift.py
krcmarik: Skipped: KubeVirt API only supports BIOS and EFI bootloader types — no other types exist in the Bootloader struct.

libs/providers/vmware.py
krcmarik: Skipped: Different signatures, one-use blocks, no dedup benefit. Doesn't solve vm_dict length.

libs/providers/openshift.py
myakove, line 215: ⚠️ **[PREV-UNRESOLVED] Non-EFI bootloader silently classified as BIOS**

If `bootloader` exists but has no `efi` key (e.g., a future KubeVirt bootloader type), the code falls to `else` and reports `"bios"`. Your response ("KubeVirt only supports BIOS and EFI") is reasonable — if skipping intentionally, please resolve the thread.

---

libs/providers/vmware.py
myakove, line 733: 💡 **[PREV-UNRESOLVED] Firmware extraction inlined in oversized `vm_dict`**

Both `vm_dict` methods exceed the 50-line guideline. Your response ("Different signatures, no dedup benefit") is noted — if skipping intentionally, please resolve the thread.

---

utilities/post_migration.py
```diff
@@ -1257,6 +1257,39 @@ def check_serial_preservation(
...
+
+    Args:
+        source_vm: Source VM info dictionary with firmware data.
+        destination_vm: Destination VM info dictionary with firmware data.
+
+    Raises:
+        AssertionError: If firmware type, secure boot, or TPM settings don't match.
+        KeyError: If required firmware keys are missing in source or destination VM info.
+    """
+    source_firmware: dict[str, Any] = source_vm["firmware"]
```
myakove, line 1379: **`check_boot_configuration` KeyError on empty firmware dict from `VIRTUAL_MACHINE_TEMPLATE`**

`base_provider.py` now defaults `"firmware": {}` in `VIRTUAL_MACHINE_TEMPLATE`. If a provider copies this template but doesn't populate firmware sub-keys (e.g., older cached result, or a provider subclass that skips firmware extraction), `source_firmware[key]` raises `KeyError` on `boot_firmware`/`secure_boot`/`tpm_present` with no VM name or context.

Per "Fail Fast - Validate Content Not Just Existence":
```suggestion
    source_firmware: dict[str, Any] = source_vm["firmware"]
    dest_firmware: dict[str, Any] = destination_vm["firmware"]

    required_keys = {"boot_firmware", "secure_boot", "tpm_present"}
    for label, firmware in [("source", source_firmware), ("destination", dest_firmware)]:
        missing = required_keys - firmware.keys()
        if missing:
            raise KeyError(
                f"Firmware data incomplete for {label} VM '{destination_vm['name']}': missing keys {missing}"
            )
```

---

libs/providers/openshift.py
```diff
@@ -192,8 +193,32 @@ def vm_dict(
         )
 
         # Serial number (from firmware) - for serial preservation verification
-        firmware_spec: dict[str, Any] | None = cnv_vm.instance.spec.template.spec.domain.get("firmware")
+        domain: dict[str, Any] = cnv_vm.instance.spec.template.spec.domain
+
+        firmware_spec: dict[str, Any] | None = domain.get("firmware")
         result_vm_info["serial"] = firmware_spec.get("serial") if firmware_spec else None
```
myakove, line 199: 💡 **OCP firmware extraction block (~25 lines) inline in already 220+ line `vm_dict`**

Same structural concern as the VMware-side comment (existing unresolved #2), but independently applicable to the OCP provider. The firmware extraction (lines 199-224) is a self-contained sub-task with clear input/output boundaries.

Suggestion: extract into `_extract_firmware_info(self, domain: dict[str, Any], cnv_vm_name: str) -> dict[str, Any]`.

---

libs/providers/vmware.py
```diff
@@ -729,6 +729,28 @@ def vm_dict(self, **kwargs: Any) -> dict[str, Any]:
...
+        firmware_info: dict[str, Any] = {}
+        boot_firmware_raw: str | None = vm_config.firmware
+        if not boot_firmware_raw:
```
myakove, line 736: 💡 **`if not boot_firmware_raw` conflates None and empty string**

Both are invalid, but the error message says "missing or empty" without distinguishing which case occurred. For clearer debugging:
```suggestion
        if boot_firmware_raw is None:
            raise ValueError(f"firmware attribute is missing for VM '{_vm.name}'")
        boot_firmware: str = boot_firmware_raw.lower()
        if not boot_firmware:
            raise ValueError(f"firmware attribute is empty for VM '{_vm.name}'")
```

---

myakove (commented):
## Code Review

### :bulb: Suggestions (5)

| File | Line | Issue |
|------|------|-------|
| `libs/providers/openshift.py` | 215 | ⚠️ **[PREV-UNRESOLVED] Non-EFI bootloader silently classified as BIOS** |
| `libs/providers/vmware.py` | 732 | 💡 **[PREV-UNRESOLVED] Firmware extraction inlined in oversized `vm_dict`** |
| `utilities/post_migration.py` | 1271 | **`check_boot_configuration` KeyError on empty firmware dict from `VIRTUAL_MACHI |
| `libs/providers/openshift.py` | 199 | 💡 **OCP firmware extraction block (~25 lines) inline in already 220+ line `vm_di |
| `libs/providers/vmware.py` | 735 | 💡 **`if not boot_firmware_raw` conflates None and empty string** |

---
*Review generated by pi*

---

libs/providers/openshift.py
krcmarik: ACK, resolving.

libs/providers/vmware.py
krcmarik: ACK, resolving.

libs/providers/openshift.py
krcmarik: ACK, resolving.

libs/providers/vmware.py
krcmarik: ACK, resolving.

utilities/post_migration.py
krcmarik: Skipping — firmware dict is our own controlled data populated by vm_dict methods in both providers. Per "Deterministic Tests" rule, bracket access is correct for data we control: a raw KeyError already fails fast and identifies the missing key. The scenario described (provider not populating firmware) would be a provider bug, not a runtime data variation. Adding pre-validation adds code without changing the fail-fast outcome.

libs/providers/openshift.py
krcmarik: Skipping — same reasoning as the VMware-side comment (resolved above): different signatures, one-use blocks, no dedup benefit. Extracting doesn't meaningfully reduce vm_dict complexity.

libs/providers/vmware.py
krcmarik: Skipping — vSphere vm.config.firmware always returns a non-empty string ("efi" or "bios"). Empty string is not a realistic scenario. Splitting the check adds two ValueError paths for a case that doesn't occur in practice, while the current single check already fails fast with a clear message.

utilities/post_migration.py
```diff
@@ -1257,6 +1257,39 @@ def check_serial_preservation(
...
+    dest_firmware: dict[str, Any] = destination_vm["firmware"]
+
+    firmware_checks: list[tuple[str, str]] = [
```
myakove, line 1382: 💡 **`firmware_checks` is a static list recreated every call — extract to module-level constant**

Per "Magic Numbers and Hardcoded Values (SHOULD)" — this is a fixed mapping that doesn't change between calls.

```suggestion
```

At module level (e.g., after `_LUKS_FSTYPE`):
```python
_FIRMWARE_CHECK_KEYS: tuple[tuple[str, str], ...] = (
    ("boot_firmware", "Boot firmware"),
    ("secure_boot", "Secure boot"),
    ("tpm_present", "TPM"),
)
```

Then in the function:
```python
    for key, label in _FIRMWARE_CHECK_KEYS:
```

---

myakove (commented):
## Code Review

### :bulb: Suggestions (2)

| File | Line | Issue |
|------|------|-------|
| `utilities/post_migration.py` | 1274 | 💡 **`firmware_checks` is a static list recreated every call — extract to module- |
| `libs/providers/vmware.py` | 741 | 💡 **`boot_options` fetched unconditionally but only used in EFI branch** |

---
*Review generated by pi*

---

libs/providers/vmware.py
```diff
@@ -729,6 +729,28 @@ def vm_dict(self, **kwargs: Any) -> dict[str, Any]:
...
+            raise ValueError(f"firmware attribute is missing or empty for VM '{_vm.name}'")
+        boot_firmware: str = boot_firmware_raw.lower()
+        if boot_firmware not in ("bios", "efi"):
+            raise ValueError(f"Unexpected firmware type '{boot_firmware}' for VM '{_vm.name}'")
+        firmware_info["boot_firmware"] = boot_firmware
+        boot_options: vim.vm.BootOptions | None = vm_config.bootOptions
```
myakove, line 742: 💡 **`boot_options` fetched unconditionally but only used in EFI branch**

Narrow the scope — move inside the conditional:
```suggestion
        if boot_firmware == "efi":
            boot_options: vim.vm.BootOptions | None = vm_config.bootOptions
            if boot_options is None:
                raise ValueError(f"bootOptions is None for EFI VM '{_vm.name}' — expected boot configuration")
            firmware_info["secure_boot"] = boot_options.efiSecureBootEnabled
```

---

krcmarik:
@coderabbitai — Regarding the two comments from your latest review (review 4620904319):

**1. CPU feature lookup re-traversing `cnv_vm.instance.spec.template.spec.domain` (openshift.py:311-315)**
Not relevant to this PR — the CPU feature extraction code at those lines is pre-existing code not modified by this PR's changes.

**2. `check_vbs_status` duplicating `run_cmd_in_vm` flow (post_migration.py:812-862)**
Not relevant to this PR — `check_vbs_status` is pre-existing code not modified by this PR. The refactoring suggestion may be valid but is out of scope.

Review dismissed.

redhat-qe-bot1:
**Clean rebase detected** — no code changes compared to previous head (`1313e42`).
The following labels were preserved: `commented-coderabbitai[bot]`, `changes-requested-coderabbitai[bot]`.

krcmarik:
@coderabbitai full review

redhat-qe-bot2:

PR #574, opened 2026-07-02
fix(warm): restrict ClusterRole tests to vSphere provider only

AmenB:
## Summary
- Add `@pytest.mark.vsphere` to both ClusterRole test classes
- The SCC requirement for `forklift-migrator-role` is vSphere-specific — on RHV the migration succeeds without SCC, causing the negative test to incorrectly fail
- The existing `pytest_collection_modifyitems` hook auto-skips items with the `vsphere` keyword for non-vSphere providers

## Summary by CodeRabbit

* **Tests**
  * Added vSphere test coverage to two warm cluster role migration test scenarios, including both the SCC and non-SCC variants.


krcmarik (approved):

rh-bot-1:

PR #583, opened 2026-07-06
fix: support vSphere template cloning with configurable resource pool

tshefi:
Add priority-based resource pool selection for VM cloning operations to support cloning from vSphere templates. Resolves DVS port conflicts by enabling template usage as clone sources.

Priority order:
1. Configured copyoffload.resource_pool (explicit override)
2. Target ESXi host's pool (automatic compatibility)
3. Source VM's pool (preserve original location)
4. Cluster-wide pool search (fallback for templates)

Related: MTV-5787

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved VM cloning by refining destination host and resource-pool selection, honoring configured options and providing safer fallbacks.
  * Enhanced network mapping during cloning by preferring existing inventory mappings first, then discovering networks directly when necessary.
  * Added clearer handling when expected network hardware/backing details are missing, reducing unexpected failures when no networks are found.
* **Tests**
  * Updated the Multus network fixture to base NAD naming on the prepared clone plan, aligning it with cloned VM/template names.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -811,6 +811,7 @@ def multus_network_name(
     source_provider: BaseProvider,
     source_provider_inventory: ForkliftInventory,
     class_plan_config: dict[str, Any],
+    prepared_plan: dict[str, Any],
     request: pytest.FixtureRequest,
 ) -> dict[str, str]:
     """Create NADs based on network requirements with unique names per test class.
@@ -830,6 +831,7 @@ def multus_network_name(
         source_provider (BaseProvider): Source provider instance
         source_provider_inventory (ForkliftInventory): Source provider inventory
         class_plan_config (dict[str, Any]): Plan configuration from class parametrization
+        prepared_plan (dict[str, Any]): Prepared plan with cloned VM names
         request (pytest.FixtureRequest): Pytest fixture request
 
     Returns:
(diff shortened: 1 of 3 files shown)


redhat-qe-bot1:

conftest.py
```diff
@@ -853,9 +855,10 @@ def multus_network_name(
     else:
         nad_namespace = target_namespace
 
-    # Get VM/template names from the class plan config
-    vms = [vm["name"] for vm in class_plan_config["virtual_machines"]]
-    LOGGER.info(f"Found VMs from class config: {vms}")
+    # Get cloned VM names from prepared_plan (not original template names from class_plan_config)
+    # After cloning, prepared_plan has the actual VM names (e.g., auto-zzsx-xcopy-template-test-...)
+    vms = [vm["name"] for vm in prepared_plan["virtual_machines"]]
+    LOGGER.info(f"Found VMs from prepared plan: {vms}")
```
krcmarik, line 861: We have other providers too, Once the VM is cloned the test proceeds to create the Storage and Network Maps I am afraid the cloned VM may not be yet  populated in the forklift inventory - in the case of vsphere it will fall back to the VMware API (with this PR) but we have other providers where It may fail, the original VM has the same networks as the cloned one so maybe It's not necessary to use the cloned one for getting the networks. We do not really add any networks to the cloned VMs anywhere in the code. I need to test this.

PR #589, opened 2026-07-07
feat(MTV-5979): add cancel and validation tests for standalone DI

MiriSafra:
## Summary

Adds two new test classes for standalone Deep Inspection (Conversion CR), extending the happy-path coverage from PR #575:

- **TestStandaloneDICancel** (8 tests, scenario 17): Creates a DI Conversion CR, waits for Running phase, verifies pods exist, cancels via status subresource patch (mirroring forklift's `CancelConversion()` pattern), verifies cleanup (pod deleted, `forklift-deep-inspection` snapshot removed by controller), then re-runs DI on the same VM and verifies success.

- **TestStandaloneDIValidation** (2 tests, scenario 18): Creates a Conversion CR without `vddkImage` to trigger controller validation. Verifies the CR reaches a non-running phase with a Critical `VDDKImageNotSet` condition.

### New utilities (`utilities/deep_inspection.py`)
- `cancel_conversion()` — patches status subresource to set `phase=Canceled, stage=Finished`
- `wait_for_conversion_phase()` — polls until a specific phase is reached, with terminal phase guard
- `wait_for_di_snapshot()` — polls vSphere for `forklift-deep-inspection` snapshot creation
- `wait_for_conversion_pods_cleanup()` — polls until conversion pods are cleaned up after cancel
- `wait_for_di_snapshot_cleanup()` — polls until DI snapshot is removed from source VM
- `wait_for_critical_conditions()` — polls for Critical conditions on a Conversion CR

### New fixtures (`tests/deep_inspection/conftest.py`)
- `vmware_source_provider` — type-validated VMWareProvider for snapshot API access
- `di_invalid_conversion_resource` — creates a Conversion CR without `vddkImage` for validation testing

## Jira
[MTV-6046](https://redhat.atlassian.net/browse/MTV-6046)

## Summary by CodeRabbit

* **New Features**
  * Enhanced standalone vSphere Deep Inspection workflows with lifecycle helpers for phase polling, conversion cancellation, DI snapshot discovery, and pod/snapshot cleanup waiting.
  * Deep Inspection conversion creation now supports omitting `vddkImage` to intentionally exercise validation behavior with standardized timeouts.
* **Tests**
  * Added standalone coverage for canceling an in-progress conversion and verifying cleanup/result re-run.
  * Added validation tests ensuring missing `vddkImage` triggers the expected critical condition.
  * Improved Deep Inspection test fixtures by pre-resolving the target VM and required inputs.
* **Bug Fixes**
  * Improved conversion completion/critical-condition detection for more accurate terminal outcomes and clearer failure details.

### DIFF
--- a/tests/deep_inspection/conftest.py
+++ b/tests/deep_inspection/conftest.py
@@ -7,7 +7,12 @@ import pytest
 from ocp_resources.conversion import Conversion
 from ocp_resources.secret import Secret
 
-from utilities.deep_inspection import create_conversion_resource, create_di_connection_secret
+from utilities.deep_inspection import (
+    CONVERSION_TYPE_DEEP_INSPECTION,
+    create_conversion_resource,
+    create_di_connection_secret,
+)
+from utilities.resources import create_and_store_resource
 from utilities.utils import populate_vm_ids
 
 if TYPE_CHECKING:
@@ -46,27 +51,64 @@ def di_connection_secret(
 
 
 @pytest.fixture(scope="class")
-def di_vm_name(class_plan_config: dict[str, Any]) -> str:
-    """First VM name from the plan config.
+def di_resolved_vm(
+    class_plan_config: dict[str, Any],
+    source_provider_inventory: "ForkliftInventory",
+) -> dict[str, str]:
+    """First VM from the plan config with populated inventory ID.
 
     Args:
         class_plan_config (dict[str, Any]): Raw plan config from parametrization.
+        source_provider_inventory (ForkliftInventory): Source provider inventory.
+
+    Returns:
+        dict[str, str]: VM dict with 'name' and 'id' keys.
+    """
+    plan_config = deepcopy(class_plan_config)
+    populate_vm_ids(plan_config, source_provider_inventory)
+    return plan_config["virtual_machines"][0]
+
+
+@pytest.fixture(scope="class")
+def di_vm_name(di_resolved_vm: dict[str, str]) -> str:
+    """First VM name from the resolved plan config.
+
+    Args:
+        di_resolved_vm (dict[str, str]): Resolved VM dict with 'name' and 'id'.
 
     Returns:
         str: The VM name.
     """
-    return class_plan_config["virtual_machines"][0]["name"]
+    return di_resolved_vm["name"]
+
+
+@pytest.fixture(scope="class")
+def di_vddk_image(source_provider_data: dict[str, Any]) -> str:
+    """VDDK init container image from provider config.
+
+    Args:
+        source_provider_data (dict[str, Any]): Provider config from providers.json.
+
+    Returns:
+        str: The VDDK init container image.
+
+    Raises:
+        ValueError: If vddk_init_image is missing from provider data.
+    """
+    vddk_image = source_provider_data.get("vddk_init_image")
+    if not vddk_image:
+        raise ValueError("vddk_init_image missing from source_provider_data — required for DeepInspection")
+    return vddk_image
 
 
 @pytest.fixture(scope="class")
 def di_conversion_resource(
-    class_plan_config: dict[str, Any],
+    di_resolved_vm: dict[str, str],
     di_connection_secret: Secret,
+    di_vddk_image: str,
     fixture_store: dict[str, Any],
     ocp_admin_client: "DynamicClient",
     target_namespace: str,
-    source_provider_data: dict[str, Any],
-    source_provider_inventory: "ForkliftInventory",
 ) -> Conversion:
     """Standalone DeepInspection Conversion CR for the first VM in the plan.
 
@@ -75,34 +117,58 @@ def di_conversion_resource(
     is needed.
 
     Args:
-        class_plan_config (dict[str, Any]): Raw plan config from parametrization.
+        di_resolved_vm (dict[str, str]): Resolved VM dict with 'name' and 'id'.
         di_connection_secret (Secret): Connection secret for vSphere access.
+        di_vddk_image (str): VDDK init container image.
         fixture_store (dict[str, Any]): Resource tracking dictionary.
         ocp_admin_client (DynamicClient): OpenShift admin client.
         target_namespace (str): Namespace for the Conversion CR.
-        source_provider_data (dict[str, Any]): Provider config from providers.json.
-        source_provider_inventory (ForkliftInventory): Source provider inventory.
 
     Returns:
         Conversion: The created Conversion CR.
-
-    Raises:
-        ValueError: If vddk_init_image is missing from provider data.
     """
-    plan_config = deepcopy(class_plan_config)
-    populate_vm_ids(plan_config, source_provider_inventory)
-    vm = plan_config["virtual_machines"][0]
-
-    vddk_image = source_provider_data.get("vddk_init_image")
-    if not vddk_image:
-        raise ValueError("vddk_init_image missing from source_provider_data — required for DeepInspection")
-
     return create_conversion_resource(
         client=ocp_admin_client,
         fixture_store=fixture_store,
         connection_secret=di_connection_secret,
-        vm_id=vm["id"],
-        vm_name=vm["name"],
-        vddk_image=vddk_image,
+        vm_id=di_resolved_vm["id"],
+        vm_name=di_resolved_vm["name"],
+        vddk_image=di_vddk_image,
+        target_namespace=target_namespace,
+    )
+
+
+@pytest.fixture(scope="class")
+def di_invalid_conversion_resource(
+    di_resolved_vm: dict[str, str],
+    di_connection_secret: Secret,
+    fixture_store: dict[str, Any],
+    ocp_admin_client: "DynamicClient",
+    target_namespace: str,
+) -> Conversion:
+    """Conversion CR with missing vddkImage to trigger validation error.
+
+    Creates a DeepInspection Conversion CR without the required vddkImage
+    field. The controller's validateVDDKImage() sets a Critical condition
+    and blocks pipeline execution.
+
+    Args:
+        di_resolved_vm (dict[str, str]): Resolved VM dict with 'name' and 'id'.
+        di_connection_secret (Secret): Connection secret for vSphere access.
+        fixture_store (dict[str, Any]): Resource tracking dictionary.
+        ocp_admin_client (DynamicClient): OpenShift admin client.
+        target_namespace (str): Namespace for the Conversion CR.
+
+    Returns:
+        Conversion: The created Conversion CR (will have Critical conditions).
+    """
+    return create_and_store_resource(
+        client=ocp_admin_client,
+        fixture_store=fixture_store,
+        resource=Conversion,
+        namespace=target_namespace,
+        type=CONVERSION_TYPE_DEEP_INSPECTION,
+        connection={"secret": {"name": di_connection_secret.name, "namespace": di_connection_secret.namespace}},
+        vm={"id": di_resolved_vm["id"], "name": di_resolved_vm["name"], "type": "VirtualMachine"},
         target_namespace=target_namespace,
     )
--- a/tests/deep_inspection/conftest.py
+++ b/tests/deep_inspection/conftest.py
@@ -8,11 +8,9 @@ from ocp_resources.conversion import Conversion
 from ocp_resources.secret import Secret
 
 from utilities.deep_inspection import (
-    CONVERSION_TYPE_DEEP_INSPECTION,
     create_conversion_resource,
     create_di_connection_secret,
 )
-from utilities.resources import create_and_store_resource
 from utilities.utils import populate_vm_ids
 
 if TYPE_CHECKING:
@@ -162,13 +160,11 @@ def di_invalid_conversion_resource(
     Returns:
         Conversion: The created Conversion CR (will have Critical conditions).
     """
-    return create_and_store_resource(
+    return create_conversion_resource(
         client=ocp_admin_client,
         fixture_store=fixture_store,
-        resource=Conversion,
-        namespace=target_namespace,
-        type=CONVERSION_TYPE_DEEP_INSPECTION,
-        connection={"secret": {"name": di_connection_secret.name, "namespace": di_connection_secret.namespace}},
-        vm={"id": di_resolved_vm["id"], "name": di_resolved_vm["name"], "type": "VirtualMachine"},
+        connection_secret=di_connection_secret,
+        vm_id=di_resolved_vm["id"],
+        vm_name=di_resolved_vm["name"],
         target_namespace=target_namespace,
     )
--- a/tests/deep_inspection/conftest.py
+++ b/tests/deep_inspection/conftest.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-from copy import deepcopy
 from typing import TYPE_CHECKING, Any
 
 import pytest
@@ -52,7 +51,7 @@ def di_connection_secret(
 def di_resolved_vm(
     class_plan_config: dict[str, Any],
     source_provider_inventory: "ForkliftInventory",
-) -> dict[str, str]:
+) -> dict[str, Any]:
     """First VM from the plan config with populated inventory ID.
 
     Args:
@@ -60,24 +59,23 @@ def di_resolved_vm(
         source_provider_inventory (ForkliftInventory): Source provider inventory.
 
     Returns:
-        dict[str, str]: VM dict with 'name' and 'id' keys.
+        dict[str, Any]: VM dict with 'name' and 'id' keys.
     """
-    plan_config = deepcopy(class_plan_config)
-    populate_vm_ids(plan_config, source_provider_inventory)
-    return plan_config["virtual_machines"][0]
+    populate_vm_ids(class_plan_config, source_provider_inventory)
+    return class_plan_config["virtual_machines"][0]
 
 
 @pytest.fixture(scope="class")
-def di_vm_name(di_resolved_vm: dict[str, str]) -> str:
-    """First VM name from the resolved plan config.
+def di_vm_name(class_plan_config: dict[str, Any]) -> str:
+    """First VM name from the plan config.
 
     Args:
-        di_resolved_vm (dict[str, str]): Resolved VM dict with 'name' and 'id'.
+        class_plan_config (dict[str, Any]): Raw plan config from parametrization.
 
     Returns:
         str: The VM name.
     """
-    return di_resolved_vm["name"]
+    return class_plan_config["virtual_machines"][0]["name"]
 
 
 @pytest.fixture(scope="class")
@@ -101,7 +99,7 @@ def di_vddk_image(source_provider_data: dict[str, Any]) -> str:
 
 @pytest.fixture(scope="class")
 def di_conversion_resource(
-    di_resolved_vm: dict[str, str],
+    di_resolved_vm: dict[str, Any],
     di_connection_secret: Secret,
     di_vddk_image: str,
     fixture_store: dict[str, Any],
@@ -115,7 +113,7 @@ def di_conversion_resource(
     is needed.
 
     Args:
-        di_resolved_vm (dict[str, str]): Resolved VM dict with 'name' and 'id'.
+        di_resolved_vm (dict[str, Any]): Resolved VM dict with 'name' and 'id'.
         di_connection_secret (Secret): Connection secret for vSphere access.
         di_vddk_image (str): VDDK init container image.
         fixture_store (dict[str, Any]): Resource tracking dictionary.
@@ -138,7 +136,7 @@ def di_conversion_resource(
 
 @pytest.fixture(scope="class")
 def di_invalid_conversion_resource(
-    di_resolved_vm: dict[str, str],
+    di_resolved_vm: dict[str, Any],
     di_connection_secret: Secret,
     fixture_store: dict[str, Any],
     ocp_admin_client: "DynamicClient",
@@ -151,7 +149,7 @@ def di_invalid_conversion_resource(
     and blocks pipeline execution.
 
     Args:
-        di_resolved_vm (dict[str, str]): Resolved VM dict with 'name' and 'id'.
+        di_resolved_vm (dict[str, Any]): Resolved VM dict with 'name' and 'id'.
         di_connection_secret (Secret): Connection secret for vSphere access.
         fixture_store (dict[str, Any]): Resource tracking dictionary.
         ocp_admin_client (DynamicClient): OpenShift admin client.
--- a/tests/deep_inspection/conftest.py
+++ b/tests/deep_inspection/conftest.py
@@ -6,6 +6,7 @@ import pytest
 from ocp_resources.conversion import Conversion
 from ocp_resources.secret import Secret
 
+from libs.providers.vmware import VMWareProvider
 from utilities.deep_inspection import (
     create_conversion_resource,
     create_di_connection_secret,
@@ -48,6 +49,27 @@ def di_connection_secret(
 
 
 @pytest.fixture(scope="class")
+def vmware_source_provider(source_provider: "BaseProvider") -> VMWareProvider:
+    """Source provider validated as VMWare type.
+
+    Snapshot operations (create, list, cleanup) require vSphere-specific
+    APIs only available on VMWareProvider.
+
+    Args:
+        source_provider (BaseProvider): Source provider instance.
+
+    Returns:
+        VMWareProvider: The source provider, type-validated.
+
+    Raises:
+        TypeError: If source provider is not a VMWareProvider.
+    """
+    if not isinstance(source_provider, VMWareProvider):
+        raise TypeError(f"Snapshot operations require VMWareProvider, got {type(source_provider).__name__}")
+    return source_provider
+
+
+@pytest.fixture(scope="class")
 def di_resolved_vm(
     class_plan_config: dict[str, Any],
     source_provider_inventory: "ForkliftInventory",
@@ -59,7 +81,7 @@ def di_resolved_vm(
         source_provider_inventory (ForkliftInventory): Source provider inventory.
 
     Returns:
-        dict[str, Any]: VM dict with 'name' and 'id' keys.
+        dict[str, Any]: VM config dict with populated 'name' and 'id' keys.
     """
     populate_vm_ids(class_plan_config, source_provider_inventory)
     return class_plan_config["virtual_machines"][0]
@@ -106,14 +128,13 @@ def di_conversion_resource(
     ocp_admin_client: "DynamicClient",
     target_namespace: str,
 ) -> Conversion:
-    """Standalone DeepInspection Conversion CR for the first VM in the plan.
+    """Standalone DeepInspection Conversion CR for the first VM.
 
-    Uses class_plan_config directly (not prepared_plan) because Deep Inspection
-    only creates a snapshot, inspects, and removes the snapshot — no VM cloning
-    is needed.
+    Bypasses prepared_plan because Deep Inspection only creates a snapshot,
+    inspects, and removes it — no VM cloning is needed.
 
     Args:
-        di_resolved_vm (dict[str, Any]): Resolved VM dict with 'name' and 'id'.
+        di_resolved_vm (dict[str, Any]): VM config dict with 'name' and 'id'.
         di_connection_secret (Secret): Connection secret for vSphere access.
         di_vddk_image (str): VDDK init container image.
         fixture_store (dict[str, Any]): Resource tracking dictionary.
--- a/tests/deep_inspection/test_standalone_di.py
+++ b/tests/deep_inspection/test_standalone_di.py
@@ -4,7 +4,6 @@ from typing import TYPE_CHECKING, Any
 
 import pytest
 from ocp_resources.conversion import Conversion
-from ocp_resources.pod import Pod
 from ocp_resources.secret import Secret
 from pytest_testconfig import config as py_config
 
@@ -16,6 +15,7 @@ from utilities.deep_inspection import (
     verify_di_results,
     wait_for_conversion_complete,
     wait_for_conversion_phase,
+    wait_for_conversion_pods,
     wait_for_conversion_pods_cleanup,
     wait_for_critical_conditions,
     wait_for_di_snapshot,
@@ -111,16 +111,9 @@ class TestStandaloneDICancel:
         )
 
     def test_cancel_conversion(self, di_vm_name: str, vmware_source_provider: "VMWareProvider") -> None:
-        """Wait for DI snapshot on source VM, then cancel the conversion."""
+        """Wait for DI pod to be running on the cluster, then cancel the conversion."""
         wait_for_di_snapshot(source_provider=vmware_source_provider, vm_name=di_vm_name)
-        pods = list(
-            Pod.get(
-                client=self.conversion.client,
-                namespace=self.conversion.namespace,
-                label_selector=f"conversion={self.conversion.name}",
-            )
-        )
-        assert pods, f"No conversion pods found before cancel for '{self.conversion.name}'"
+        wait_for_conversion_pods(conversion=self.conversion)
         cancel_conversion(conversion=self.conversion)
 
     @pytest.mark.usefixtures("class_plan_config")
--- a/tests/deep_inspection/test_standalone_di.py
+++ b/tests/deep_inspection/test_standalone_di.py
@@ -1,10 +1,31 @@
 from __future__ import annotations
 
+from typing import TYPE_CHECKING, Any
+
 import pytest
 from ocp_resources.conversion import Conversion
+from ocp_resources.pod import Pod
+from ocp_resources.secret import Secret
 from pytest_testconfig import config as py_config
+from timeout_sampler import TimeoutExpiredError, TimeoutSampler
 
-from utilities.deep_inspection import verify_di_results, wait_for_conversion_complete
+from exceptions.exceptions import ConversionError
+from utilities.deep_inspection import (
+    DI_POD_CLEANUP_TIMEOUT,
+    DI_SNAPSHOT_CLEANUP_TIMEOUT,
+    DI_SNAPSHOT_NAME,
+    VDDK_IMAGE_NOT_SET_CONDITION,
+    cancel_conversion,
+    create_conversion_resource,
+    verify_di_results,
+    wait_for_conversion_complete,
+    wait_for_conversion_phase,
+    wait_for_di_snapshot,
+)
+
+if TYPE_CHECKING:
+    from kubernetes.dynamic import DynamicClient
+    from libs.providers.vmware import VMWareProvider
 
 
 @pytest.mark.parametrize(
@@ -49,3 +70,202 @@ class TestStandaloneDeepInspection:
     def test_verify_di_results(self, di_vm_name: str) -> None:
         """Verify inspection results contain OS info and filesystem data."""
         verify_di_results(conversion=self.conversion, vm_name=di_vm_name)
+
+
+@pytest.mark.parametrize(
+    "class_plan_config",
+    [
+        pytest.param(
+            py_config["tests_params"]["test_standalone_di_vsphere"],
+        ),
+    ],
+    indirect=True,
+    ids=["standalone-di-cancel-vsphere"],
+)
+@pytest.mark.incremental
+@pytest.mark.tier1
+@pytest.mark.deep_inspection
+@pytest.mark.vsphere
+class TestStandaloneDICancel:
+    """Cancel a running standalone Deep Inspection and re-run.
+
+    Verifies the cancel mechanism: patch status.phase to Canceled,
+    controller deletes the pod and cleans up snapshots. Then creates
+    a new DI for the same VM and verifies it succeeds.
+    """
+
+    conversion: Conversion
+    rerun_conversion: Conversion
+
+    def test_create_and_cancel_conversion(
+        self,
+        di_conversion_resource: Conversion,
+        di_vm_name: str,
+        source_provider: "VMWareProvider",
+    ) -> None:
+        """Create DI Conversion, wait for Running and snapshot, then cancel."""
+        self.__class__.conversion = di_conversion_resource
+        wait_for_conversion_phase(
+            conversion=self.conversion,
+            phase=Conversion.Status.RUNNING,
+            timeout=py_config["plan_wait_timeout"],
+        )
+        wait_for_di_snapshot(source_provider=source_provider, vm_name=di_vm_name)
+        cancel_conversion(conversion=self.conversion)
+
+    @pytest.mark.usefixtures("di_vm_name")
+    def test_verify_cancel_cleanup(self) -> None:
+        """Verify canceled conversion reaches Canceled phase and pod is cleaned up."""
+        with pytest.raises(ConversionError, match="Canceled"):
+            wait_for_conversion_complete(
+                conversion=self.conversion,
+                timeout=py_config["plan_wait_timeout"],
+            )
+
+        try:
+            for pods in TimeoutSampler(
+                wait_timeout=DI_POD_CLEANUP_TIMEOUT,
+                sleep=5,
+                func=lambda: list(
+                    Pod.get(
+                        client=self.conversion.client,
+                        namespace=self.conversion.namespace,
+                        label_selector=f"conversion={self.conversion.name}",
+                    )
+                ),
+            ):
+                if not pods:
+                    return
+        except TimeoutExpiredError:
+            remaining = list(
+                Pod.get(
+                    client=self.conversion.client,
+                    namespace=self.conversion.namespace,
+                    label_selector=f"conversion={self.conversion.name}",
+                )
+            )
+            raise AssertionError(
+                f"Conversion pods still present {DI_POD_CLEANUP_TIMEOUT}s after cancel: {[p.name for p in remaining]}"
+            )
+
+    def test_verify_snapshot_cleanup(self, di_vm_name: str, source_provider: "VMWareProvider") -> None:
+        """Verify forklift DI snapshot is removed from the source VM after cancel."""
+        vm = source_provider.get_vm_by_name(query=di_vm_name)
+        assert vm, f"VM '{di_vm_name}' not found in source provider"
+        try:
+            for snapshots in TimeoutSampler(
+                wait_timeout=DI_SNAPSHOT_CLEANUP_TIMEOUT,
+                sleep=10,
+                func=lambda: [s for s in source_provider.list_snapshots(vm) if s.name == DI_SNAPSHOT_NAME],
+            ):
+                if not snapshots:
+                    return
+        except TimeoutExpiredError:
+            remaining = [s.name for s in source_provider.list_snapshots(vm)]
+            raise AssertionError(
+                f"DI snapshot '{DI_SNAPSHOT_NAME}' still present on VM '{di_vm_name}' "
+                f"{DI_SNAPSHOT_CLEANUP_TIMEOUT}s after cancel. All snapshots: {remaining}"
+            )
+
+    def test_rerun_di_after_cancel(
+        self,
+        di_resolved_vm: dict[str, str],
+        di_connection_secret: Secret,
+        di_vddk_image: str,
+        fixture_store: dict[str, Any],
+        ocp_admin_client: "DynamicClient",
+        target_namespace: str,
+    ) -> None:
+        """Create a new DI Conversion for the same VM after cancel."""
+        self.__class__.rerun_conversion = create_conversion_resource(
+            client=ocp_admin_client,
+            fixture_store=fixture_store,
+            connection_secret=di_connection_secret,
+            vm_id=di_resolved_vm["id"],
+            vm_name=di_resolved_vm["name"],
+            vddk_image=di_vddk_image,
+            target_namespace=target_namespace,
+        )
+        assert self.rerun_conversion
+
+    @pytest.mark.usefixtures("di_vm_name")
+    def test_rerun_wait_for_completion(self) -> None:
+        """Wait for the re-run DI to complete successfully."""
+        wait_for_conversion_complete(
+            conversion=self.rerun_conversion,
+            timeout=py_config["plan_wait_timeout"],
+        )
+
+    def test_rerun_verify_results(self, di_vm_name: str) -> None:
+        """Verify the re-run DI produced valid inspection results."""
+        verify_di_results(conversion=self.rerun_conversion, vm_name=di_vm_name)
+
+
+@pytest.mark.parametrize(
+    "class_plan_config",
+    [
+        pytest.param(
+            py_config["tests_params"]["test_standalone_di_vsphere"],
+        ),
+    ],
+    indirect=True,
+    ids=["standalone-di-validation-vsphere"],
+)
+@pytest.mark.incremental
+@pytest.mark.tier1
+@pytest.mark.deep_inspection
+@pytest.mark.vsphere
+class TestStandaloneDIValidation:
+    """Validation errors for standalone Deep Inspection.
+
+    Creates a Conversion CR with missing vddkImage (required for
+    DeepInspection type). The controller sets a Critical condition
+    (VDDKImageNotSet) and blocks the pipeline.
+    """
+
+    conversion: Conversion
+
+    def test_create_invalid_conversion(self, di_invalid_conversion_resource: Conversion) -> None:
+        """Create a DeepInspection Conversion CR without vddkImage."""
+        self.__class__.conversion = di_invalid_conversion_resource
+        assert self.conversion
+
+    @pytest.mark.usefixtures("di_vm_name")
+    def test_verify_validation_error(self) -> None:
+        """Verify controller sets Critical condition for missing vddkImage.
+
+        Polls status until conditions appear (controller may not have
+        reconciled immediately after CR creation).
+        """
+        timeout = py_config["plan_wait_timeout"]
+        critical_conditions: list[Any] = []
+        phase = ""
+
+        try:
+            for sample in TimeoutSampler(
+                wait_timeout=timeout,
+                sleep=3,
+                func=lambda: self.conversion.instance.status,
+            ):
+                if not sample:
+                    continue
+                conditions = sample.get("conditions", [])
+                critical_conditions = [
+                    c for c in conditions if c.get("category") == "Critical" and c.get("status") == "True"
+                ]
+                if critical_conditions:
+                    phase = sample.get("phase", "")
+                    break
+        except TimeoutExpiredError:
+            raise AssertionError(f"Conversion '{self.conversion.name}' expected Critical conditions within {timeout}s")
+
+        condition_types = [c.get("type") for c in critical_conditions]
+        assert VDDK_IMAGE_NOT_SET_CONDITION in condition_types, (
+            f"Conversion '{self.conversion.name}' expected {VDDK_IMAGE_NOT_SET_CONDITION} condition, "
+            f"got: {condition_types}"
+        )
+
+        assert phase not in {Conversion.Status.RUNNING, Conversion.Status.SUCCEEDED}, (
+            f"Conversion '{self.conversion.name}' should not reach Running or Succeeded with validation errors, "
+            f"got phase '{phase}'"
+        )
--- a/tests/deep_inspection/test_standalone_di.py
+++ b/tests/deep_inspection/test_standalone_di.py
@@ -239,6 +239,7 @@ class TestStandaloneDIValidation:
         """
         timeout = py_config["plan_wait_timeout"]
         critical_conditions: list[Any] = []
+        conditions: list[Any] = []
         phase = ""
 
         try:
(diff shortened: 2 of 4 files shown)


utilities/deep_inspection.py
```diff
@@ -96,6 +107,107 @@ def create_di_connection_secret(
...
+            if current_phase == phase:
+                LOGGER.info(f"Conversion '{conversion.name}' reached phase '{phase}'")
+                return
+
+            if not is_target_terminal and current_phase in CONVERSION_TERMINAL_PHASES:
+                raise ConversionError(
+                    conversion_name=conversion.name,
+                    phase=current_phase,
+                    message=f"Reached terminal phase '{current_phase}' before target phase '{phase}'",
+                )
```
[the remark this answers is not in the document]
MiriSafra: Fixed: removed the `is_target_terminal` guard so terminal phase mismatches now raise `ConversionError` immediately regardless of whether the target phase is terminal or not.

tests/deep_inspection/test_standalone_di.py
```diff
@@ -49,3 +70,202 @@ def test_wait_for_completion(self) -> None:
...
+        self.__class__.conversion = di_conversion_resource
+        wait_for_conversion_phase(
+            conversion=self.conversion,
+            phase=Conversion.Status.RUNNING,
+            timeout=py_config["plan_wait_timeout"],
+        )
+        wait_for_di_snapshot(source_provider=source_provider, vm_name=di_vm_name)
+        cancel_conversion(conversion=self.conversion)
+
+    @pytest.mark.usefixtures("di_vm_name")
```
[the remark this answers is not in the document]
MiriSafra: Keeping `usefixtures("di_vm_name")` — it's structurally required to maintain the `indirect=True` parametrize chain. DI tests don't use `cleanup_migrated_vms`/`prepared_plan` (unlike standard migration tests), so without this decorator, test methods with no fixture depending on `class_plan_config` fail collection with "function uses no fixture 'class_plan_config'".

tests/deep_inspection/test_standalone_di.py
```diff
@@ -49,3 +70,202 @@ def test_wait_for_completion(self) -> None:
...
+            remaining = list(
+                Pod.get(
+                    client=self.conversion.client,
+                    namespace=self.conversion.namespace,
+                    label_selector=f"conversion={self.conversion.name}",
+                )
+            )
+            raise AssertionError(
+                f"Conversion pods still present {DI_POD_CLEANUP_TIMEOUT}s after cancel: {[p.name for p in remaining]}"
+            )
```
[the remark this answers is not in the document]
MiriSafra: This is pytest's actual exception class name — `pytest.raises` catches `AssertionError` not `AssertionError`. The spelling in the `raise` statement matches what pytest expects. No change needed.

tests/deep_inspection/test_standalone_di.py
```diff
@@ -49,3 +70,202 @@ def test_wait_for_completion(self) -> None:
...
+        self.__class__.rerun_conversion = create_conversion_resource(
+            client=ocp_admin_client,
+            fixture_store=fixture_store,
+            connection_secret=di_connection_secret,
+            vm_id=di_resolved_vm["id"],
+            vm_name=di_resolved_vm["name"],
+            vddk_image=di_vddk_image,
+            target_namespace=target_namespace,
+        )
+        assert self.rerun_conversion
```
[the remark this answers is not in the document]
MiriSafra: Same reason as the other `usefixtures` comment — structurally required for the `indirect=True` parametrize chain. See reply on the other thread.

utilities/deep_inspection.py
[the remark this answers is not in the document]
MiriSafra: Fixed: removed `is_target_terminal` guard. The happy-path match (`current_phase == phase`) fires first on line 161, so the terminal-mismatch check on line 165 only fires for unexpected terminal phases.

utilities/deep_inspection.py
```diff
@@ -96,6 +107,107 @@ def create_di_connection_secret(
...
+    """
+    vm = source_provider.get_vm_by_name(query=vm_name)
+    for snapshots in TimeoutSampler(
+        wait_timeout=timeout,
+        sleep=5,
+        func=lambda: [s for s in source_provider.list_snapshots(vm) if s.name == DI_SNAPSHOT_NAME],
+    ):
+        if snapshots:
+            LOGGER.info(f"DI snapshot '{DI_SNAPSHOT_NAME}' found on VM '{vm_name}'")
+            return
```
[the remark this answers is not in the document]
MiriSafra: Fixed: added `if not vm: raise ValueError(...)` after `get_vm_by_name()` call. Fails fast with a clear message instead of passing `None` to `list_snapshots()`.

utilities/deep_inspection.py
```diff
@@ -96,14 +107,116 @@ def create_di_connection_secret(
...
+        source_provider (BaseProvider): Source provider with list_snapshots method.
+        vm_name (str): Source VM name to check snapshots on.
+        timeout (int): Maximum wait time in seconds.
+
+    Raises:
+        TimeoutExpiredError: If snapshot does not appear within timeout.
+    """
+    vm = source_provider.get_vm_by_name(query=vm_name)
+    if not vm:
+        raise ValueError(f"VM '{vm_name}' not found in source provider")
```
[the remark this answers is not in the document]
MiriSafra: Fixed: replaced `if not vm:` with `if vm is None:` in `wait_for_di_snapshot()` and `if vddk_image:` with `if vddk_image is not None:` in `create_conversion_resource()` — using explicit identity checks per PEP 8. Left `if settings:` as-is since that checks dict emptiness, not None.

utilities/deep_inspection.py
```diff
@@ -96,14 +107,116 @@ def create_di_connection_secret(
...
+    if vm is None:
+        raise ValueError(f"VM '{vm_name}' not found in source provider")
+    for snapshots in TimeoutSampler(
+        wait_timeout=timeout,
+        sleep=5,
+        func=lambda: [s for s in source_provider.list_snapshots(vm) if s.name == DI_SNAPSHOT_NAME],
+    ):
+        if snapshots:
+            LOGGER.info(f"DI snapshot '{DI_SNAPSHOT_NAME}' found on VM '{vm_name}'")
+            return
```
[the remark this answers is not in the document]
MiriSafra: Fixed: wrapped `TimeoutSampler` in try/except to re-raise `TimeoutExpiredError` with VM name, snapshot name, and timeout in the message. Consistent with `wait_for_conversion_phase()` and `wait_for_conversion_complete()` which also catch and re-raise with context.

rh-bot-1:
**Clean rebase detected** — no code changes compared to previous head (`b853a91`).
The following labels were preserved: `verified`, `commented-MiriSafra`.

tests/deep_inspection/test_standalone_di.py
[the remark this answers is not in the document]
MiriSafra: The two cleanup blocks poll different resources (Pod.get with label_selector vs list_snapshots with name filter), produce different error messages, and are each used exactly once. At ~15 lines each with different internals, extracting a shared helper would add abstraction without meaningful DRY benefit. The codebase generally inlines one-off TimeoutSampler patterns in tests.

utilities/deep_inspection.py
```diff
@@ -96,14 +107,120 @@ def create_di_connection_secret(
...
+    Args:
+        conversion (Conversion): The Conversion CR to cancel.
+    """
+    api_resource = conversion.client.resources.get(api_version=conversion.api_version, kind=conversion.kind)
+    api_resource.status.patch(
+        body={"status": {"phase": CONVERSION_CANCELED_PHASE, "stage": CONVERSION_STAGE_FINISHED}},
+        name=conversion.name,
+        namespace=conversion.namespace,
+        content_type="application/merge-patch+json",
+    )
```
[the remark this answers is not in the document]
MiriSafra: Valid concern. The status subresource patch (`/status` endpoint) is not exposed by openshift-python-wrapper's `Resource` base class — it only supports main-resource reads/updates. Adding a `patch_status()` method requires an upstream PR to openshift-python-wrapper. Tracked for a follow-up; can't fix within this PR's scope.

utilities/deep_inspection.py
```diff
@@ -96,14 +107,120 @@ def create_di_connection_secret(
...
+        timeout (int): Maximum wait time in seconds.
+
+    Raises:
+        ValueError: If the source VM is not found in the provider.
+        TimeoutExpiredError: If snapshot does not appear within timeout.
+    """
+    vm = source_provider.get_vm_by_name(query=vm_name)
+    if vm is None:
+        raise ValueError(f"VM '{vm_name}' not found in source provider")
+    try:
```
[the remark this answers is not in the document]
MiriSafra: Fixed: removed the dead `if vm is None` guard. Note: `get_vm_by_name()` actually raises `ValueError` (re-raised from `get_obj` at vmware.py:276) when `clone_vm=False`, not `VmNotFoundError`. Either way, it never returns None, so the guard was dead code. The docstring's `Raises: ValueError` is correct since that's what `get_vm_by_name` propagates.

utilities/deep_inspection.py
```diff
@@ -96,14 +107,120 @@ def create_di_connection_secret(
...
+        for snapshots in TimeoutSampler(
+            wait_timeout=timeout,
+            sleep=5,
+            func=lambda: [s for s in source_provider.list_snapshots(vm) if s.name == DI_SNAPSHOT_NAME],
+        ):
+            if snapshots:
+                LOGGER.info(f"DI snapshot '{DI_SNAPSHOT_NAME}' found on VM '{vm_name}'")
+                return
+    except TimeoutExpiredError:
```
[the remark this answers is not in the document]
MiriSafra: Valid observation about `list_snapshots()` only traversing the first root snapshot chain (vmware.py:374-384). However, this is pre-existing code not modified in this PR — the limitation affects all snapshot consumers, not just DI. In practice, the DI VM starts clean (no existing snapshots), so the single-chain traversal finds the DI snapshot correctly. A fix to `list_snapshots()` to do full DFS/BFS traversal belongs in a separate PR.

tests/deep_inspection/test_standalone_di.py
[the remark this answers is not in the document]
MiriSafra: Fixed: added `from err` exception chaining to all 3 occurrences (pod cleanup at line 139, snapshot cleanup at line 163, validation timeout at line 259). Original `TimeoutExpiredError` context is now preserved in tracebacks.

utilities/deep_inspection.py
```diff
@@ -96,14 +107,120 @@ def create_di_connection_secret(
...
+                    message=f"Reached terminal phase '{current_phase}' before target phase '{phase}'",
+                )
+
+    except TimeoutExpiredError:
+        raise ConversionError(
+            conversion_name=conversion.name,
+            phase=last_phase or "Unknown",
+            message=f"Did not reach phase '{phase}' within {timeout}s",
+        )
+
```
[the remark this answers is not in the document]
MiriSafra: Fixed: added `from err` exception chaining in `wait_for_conversion_phase` (line 172) and also in `wait_for_conversion_complete` (line 310) which had the same pattern. Matches the existing convention in `create_di_connection_secret`.

utilities/deep_inspection.py
```diff
@@ -96,14 +107,120 @@ def create_di_connection_secret(
...
+            wait_timeout=timeout,
+            sleep=5,
+            func=lambda: [s for s in source_provider.list_snapshots(vm) if s.name == DI_SNAPSHOT_NAME],
+        ):
+            if snapshots:
+                LOGGER.info(f"DI snapshot '{DI_SNAPSHOT_NAME}' found on VM '{vm_name}'")
+                return
+    except TimeoutExpiredError:
+        raise TimeoutExpiredError(f"DI snapshot '{DI_SNAPSHOT_NAME}' not found on VM '{vm_name}' within {timeout}s")
+
```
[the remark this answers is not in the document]
MiriSafra: Fixed both parts: (1) Removed dead `if vm is None` guard — `get_vm_by_name()` raises `ValueError` (via `get_obj` re-raise at vmware.py:276), never returns None. Note: the actual exception is `ValueError`, not `VmNotFoundError` as stated — `VmNotFoundError` only occurs in the clone failure path. (2) Added `from err` to the `TimeoutExpiredError` handler.

tests/deep_inspection/test_standalone_di.py
```diff
@@ -49,3 +70,204 @@ def test_wait_for_completion(self) -> None:
...
+                critical_conditions = [
+                    c for c in conditions if c.get("category") == "Critical" and c.get("status") == "True"
+                ]
+                if critical_conditions:
+                    phase = sample.get("phase", "")
+                    break
+        except TimeoutExpiredError as err:
+            raise AssertionError(
+                f"Conversion '{self.conversion.name}' expected Critical conditions within {timeout}s"
+            ) from err
```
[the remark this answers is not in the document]
MiriSafra: Fixed: pre-initialized `conditions` before the loop and added last observed `phase` and `condition_types` to the timeout AssertionError message. Now a CI timeout shows e.g. `Last phase='Pending', conditions=['VDDKImageNotSet']` instead of a generic message.

tests/deep_inspection/test_standalone_di.py
```diff
@@ -49,3 +70,207 @@ def test_wait_for_completion(self) -> None:
...
+                func=lambda: self.conversion.instance.status,
+            ):
+                if not sample:
+                    continue
+                conditions = sample.get("conditions", [])
+                phase = sample.get("phase", "")
+                critical_conditions = [
+                    c for c in conditions if c.get("category") == "Critical" and c.get("status") == "True"
+                ]
+                if critical_conditions:
```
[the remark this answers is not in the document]
MiriSafra: Fixed in d9c1b84: changed `c.get("type") == "Critical"` to `c.get("category") == "Critical"` in `wait_for_conversion_complete()` at `utilities/deep_inspection.py:302`. The condition model uses `category` for severity and `type` for the condition name — verified against forklift source (`conversion.go:210`) and existing patterns in `mtv_migration.py:352-353`.

tests/deep_inspection/test_standalone_di.py
```diff
@@ -49,3 +70,207 @@ def test_wait_for_completion(self) -> None:
...
+                sleep=3,
+                func=lambda: self.conversion.instance.status,
+            ):
+                if not sample:
+                    continue
+                conditions = sample.get("conditions", [])
+                phase = sample.get("phase", "")
+                critical_conditions = [
+                    c for c in conditions if c.get("category") == "Critical" and c.get("status") == "True"
+                ]
```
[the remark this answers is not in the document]
MiriSafra: Fixed in d9c1b84: changed `c.get("type") == "Critical"` to `c.get("category") == "Critical"` in `wait_for_conversion_complete()` at `utilities/deep_inspection.py:302`. Now consistent with the test at line 256 and `mtv_migration.py:352-353`.

utilities/deep_inspection.py
```diff
@@ -15,10 +15,21 @@
...
+CONVERSION_CANCELED_PHASE = Conversion.Condition.CANCELED
+# Not exposed by openshift-python-wrapper — mirrors forklift constants:
+CONVERSION_TYPE_DEEP_INSPECTION = "DeepInspection"  # forklift: DeepInspection (conversion.go:20)
+CONVERSION_STAGE_FINISHED = "Finished"  # forklift: StageFinished (conversion.go:55)
+VDDK_IMAGE_NOT_SET_CONDITION = "VDDKImageNotSet"  # forklift: validation.go:14
+DI_SNAPSHOT_NAME = "forklift-deep-inspection"  # forklift: snapshotName (client.go:21)
+DI_SNAPSHOT_CREATION_TIMEOUT = 120
+DI_POD_CLEANUP_TIMEOUT = 60
+DI_SNAPSHOT_CLEANUP_TIMEOUT = 180
+DI_SNAPSHOT_REMOVAL_TIMEOUT = 120
```
krcmarik: `DI_SNAPSHOT_REMOVAL_TIMEOUT` is defined but never used.

`DI_SNAPSHOT_REMOVAL_TIMEOUT = 120` is declared alongside the other constants but is never referenced anywhere in this PR. The snapshot cleanup polling in `test_verify_snapshot_cleanup` uses `DI_SNAPSHOT_CLEANUP_TIMEOUT` (180s) instead. If this constant is intended for future use, it should be added when that code is written.

Suggestion: Remove `DI_SNAPSHOT_REMOVAL_TIMEOUT` from this PR.

utilities/deep_inspection.py
```diff
@@ -15,10 +15,21 @@
 if TYPE_CHECKING:
     from kubernetes.dynamic import DynamicClient
     from libs.base_provider import BaseProvider
+    from libs.providers.vmware import VMWareProvider
 
 LOGGER = get_logger(name=__name__)
 
 CONVERSION_TERMINAL_PHASES = {Conversion.Status.SUCCEEDED, Conversion.Status.FAILED, Conversion.Condition.CANCELED}
+CONVERSION_CANCELED_PHASE = Conversion.Condition.CANCELED
```
krcmarik: nit: Consider using `Conversion.Condition.CANCELED` directly instead of aliasing.

`CONVERSION_CANCELED_PHASE` is assigned `Conversion.Condition.CANCELED` and then used in `cancel_conversion()`. Since the SDK already exposes a well-named constant, the alias adds a layer of indirection without adding clarity. The other constants in this block (`CONVERSION_TYPE_DEEP_INSPECTION`, `VDDK_IMAGE_NOT_SET_CONDITION`, etc.) are justified because those strings aren't exposed by the SDK — but this one is.

Not a blocker — just a consistency note.

tests/deep_inspection/conftest.py
```diff
@@ -46,27 +49,64 @@ def di_connection_secret(
 
 
 @pytest.fixture(scope="class")
-def di_vm_name(class_plan_config: dict[str, Any]) -> str:
-    """First VM name from the plan config.
+def di_resolved_vm(
+    class_plan_config: dict[str, Any],
+    source_provider_inventory: "ForkliftInventory",
+) -> dict[str, str]:
```
krcmarik: `di_resolved_vm` return type `dict[str, str]` is too narrow.

The return annotation says `dict[str, str]` but `populate_vm_ids` adds an `id` field from the inventory, and the original VM dict can contain non-string values (`"guest_agent": True`, `"source_vm_power": "on"`). The actual type is `dict[str, Any]`. The narrow annotation will mislead type checkers and callers about what the dict actually contains.

Suggestion: Change return type to `dict[str, Any]` and update the docstring and all downstream parameter annotations (`di_vm_name`, `di_conversion_resource`, `di_invalid_conversion_resource`, `test_rerun_di_after_cancel`).

tests/deep_inspection/test_standalone_di.py
```diff
@@ -1,10 +1,31 @@
...
+    verify_di_results,
+    wait_for_conversion_complete,
+    wait_for_conversion_phase,
+    wait_for_di_snapshot,
+)
 
-from utilities.deep_inspection import verify_di_results, wait_for_conversion_complete
+if TYPE_CHECKING:
+    from kubernetes.dynamic import DynamicClient
+    from libs.providers.vmware import VMWareProvider
```
krcmarik, line 27: `source_provider` typed as `VMWareProvider` — type lie for multi-provider code.

The `TYPE_CHECKING` import pulls in `VMWareProvider` and multiple test methods annotate `source_provider: "VMWareProvider"`. While DI tests are currently vSphere-only (`@pytest.mark.vsphere`), the fixture returns `BaseProvider`. Annotating as `VMWareProvider` is a type lie — it will cause type checker false positives if DI is ever extended to other providers, and it contradicts the project pattern where `source_provider` is typed as `BaseProvider`.

Suggestion: Use `BaseProvider` annotation. If vSphere-specific methods are called (like `get_vm_by_name`, `list_snapshots`), add a runtime `isinstance` check and narrow there, per CLAUDE.md's "`hasattr` / Duck-Typing Is Fragile" rule.

tests/deep_inspection/test_standalone_di.py
```diff
@@ -49,3 +70,207 @@ def test_wait_for_completion(self) -> None:
...
+
+    Verifies the cancel mechanism: patch status.phase to Canceled,
+    controller deletes the pod and cleans up snapshots. Then creates
+    a new DI for the same VM and verifies it succeeds.
+    """
+
+    conversion: Conversion
+    rerun_conversion: Conversion
+
+    def test_create_and_cancel_conversion(
```
krcmarik: `test_create_and_cancel_conversion` — consider whether this is a test or setup.

This method bundles 4 operations: (1) stores the conversion fixture, (2) waits for Running phase, (3) waits for snapshot creation, (4) cancels. None of these are assertions of the feature under test — they're preconditions for `test_verify_cancel_cleanup` and `test_verify_snapshot_cleanup`.

If this is setup, it may belong in a fixture rather than a `test_` method. If it stays as a test step, the name is misleading for triage — if `wait_for_conversion_phase` fails, the test report says "create_and_cancel" failed, not "didn't reach Running."

Consider either:
- Moving the setup logic to a fixture (like how `di_conversion_resource` handles creation), or
- Splitting into separate incremental steps (`test_create_conversion`, `test_wait_for_running`, `test_cancel_conversion`) so each step's name tells you exactly what failed

tests/deep_inspection/test_standalone_di.py
```diff
@@ -49,3 +70,207 @@ def test_wait_for_completion(self) -> None:
...
+
+    conversion: Conversion
+
+    def test_create_invalid_conversion(self, di_invalid_conversion_resource: Conversion) -> None:
+        """Create a DeepInspection Conversion CR without vddkImage."""
+        self.__class__.conversion = di_invalid_conversion_resource
+        assert self.conversion
+
+    @pytest.mark.usefixtures("di_vm_name")
+    def test_verify_validation_error(self) -> None:
```
krcmarik, line 198: `test_verify_validation_error` contains ~35 lines of inline polling logic.

This is the same pattern used in `wait_for_conversion_phase` and `wait_for_conversion_complete` — poll status, check conditions, raise on timeout. Consider extracting to a utility function like `wait_for_critical_condition()` in `utilities/deep_inspection.py`. This would:
- Keep test methods focused on assertions rather than polling mechanics
- Be reusable if other validation error tests are added
- Follow the project pattern where polling logic lives in utilities, not test methods

tests/deep_inspection/conftest.py
```diff
@@ -46,27 +49,64 @@ def di_connection_secret(
 
 
 @pytest.fixture(scope="class")
-def di_vm_name(class_plan_config: dict[str, Any]) -> str:
-    """First VM name from the plan config.
+def di_resolved_vm(
```
krcmarik, line 73: `di_resolved_vm` adds unnecessary complexity for DI's single-VM, no-clone scenario.

This fixture does `deepcopy(class_plan_config)` → `populate_vm_ids()` → `return plan_config["virtual_machines"][0]`. DI tests use a dedicated VM (`mtv-tests-di-rhel8`) that doesn't get cloned — the deepcopy protects nothing because DI doesn't mutate the config. The `populate_vm_ids` call is a simple inventory lookup that adds a VM `id` field.

The previous code in `di_conversion_resource` did this inline in 3 lines. Extracting it into a separate fixture adds an inventory dependency to `di_vm_name` (which only needs the name — unchanged by the lookup) and introduces a layer of indirection.

The fixture is consumed by `di_conversion_resource`, `di_invalid_conversion_resource`, and `test_rerun_di_after_cancel`, so there is reuse value. The concern is mainly about the unnecessary `deepcopy` and the added inventory dependency on `di_vm_name`.

Suggestion: If keeping the fixture, remove the `deepcopy` (DI doesn't mutate config) and decouple `di_vm_name` from it — `di_vm_name` should stay as `class_plan_config["virtual_machines"][0]["name"]` since it doesn't need the inventory ID.

tests/deep_inspection/conftest.py
```diff
@@ -46,27 +49,64 @@ def di_connection_secret(
...
+    Returns:
+        dict[str, str]: VM dict with 'name' and 'id' keys.
+    """
+    plan_config = deepcopy(class_plan_config)
+    populate_vm_ids(plan_config, source_provider_inventory)
+    return plan_config["virtual_machines"][0]
+
+
+@pytest.fixture(scope="class")
+def di_vm_name(di_resolved_vm: dict[str, str]) -> str:
```
krcmarik: `di_vm_name` now has an unnecessary dependency on `di_resolved_vm`.

On `main`, `di_vm_name` was a simple one-liner: `class_plan_config["virtual_machines"][0]["name"]`. This PR changes it to depend on `di_resolved_vm`, which pulls in `source_provider_inventory` just to look up the VM ID — but `di_vm_name` only uses the `name` field, which is unchanged by the inventory lookup.

Every consumer of `di_vm_name` in this PR uses it as a plain name string: `verify_di_results(vm_name=...)`, `get_vm_by_name(query=...)`, `wait_for_di_snapshot(vm_name=...)`, and several `@pytest.mark.usefixtures("di_vm_name")` calls. None of them need the VM ID. The ID is consumed through a separate path (`di_resolved_vm["id"]` in `di_conversion_resource` and `di_invalid_conversion_resource`).

Suggestion: Keep `di_vm_name` sourced from `class_plan_config` directly, as it was on `main`.

utilities/deep_inspection.py
MiriSafra: Done — removed `DI_SNAPSHOT_REMOVAL_TIMEOUT`. It was dead code; snapshot cleanup uses `DI_SNAPSHOT_CLEANUP_TIMEOUT` (180s).

utilities/deep_inspection.py
MiriSafra: Done — removed the `CONVERSION_CANCELED_PHASE` alias. Now using `Conversion.Condition.CANCELED` directly in `CONVERSION_TERMINAL_PHASES` and `cancel_conversion()`.

tests/deep_inspection/conftest.py
MiriSafra: Done — changed return type to `dict[str, Any]` and updated all downstream annotations.

tests/deep_inspection/test_standalone_di.py
MiriSafra: Done — moved `VMWareProvider` to a top-level import, changed `source_provider` annotations to `BaseProvider`, and added `isinstance(source_provider, VMWareProvider)` runtime checks where vSphere-specific methods are called.

tests/deep_inspection/test_standalone_di.py
MiriSafra: Done — split into 3 incremental steps: `test_create_conversion`, `test_wait_for_running`, `test_cancel_conversion`. Each step's name now maps directly to what failed for better triage.

tests/deep_inspection/test_standalone_di.py
MiriSafra: Done — extracted `wait_for_critical_conditions()` to `utilities/deep_inspection.py`. Test now calls the utility and asserts on the returned conditions and phase.

tests/deep_inspection/conftest.py
MiriSafra: Done — removed deepcopy (DI doesn't clone VMs, mutation is idempotent) and decoupled `di_vm_name` to read directly from `class_plan_config`.

tests/deep_inspection/conftest.py
MiriSafra: Done — `di_vm_name` now reads directly from `class_plan_config["virtual_machines"][0]["name"]`, no dependency on `di_resolved_vm` or inventory.

tests/deep_inspection/conftest.py
```diff
@@ -45,6 +47,24 @@ def di_connection_secret(
...
+
+    Args:
+        class_plan_config (dict[str, Any]): Raw plan config from parametrization.
+        source_provider_inventory (ForkliftInventory): Source provider inventory.
+
+    Returns:
+        dict[str, Any]: VM dict with 'name' and 'id' keys.
+    """
+    populate_vm_ids(class_plan_config, source_provider_inventory)
+    return class_plan_config["virtual_machines"][0]
```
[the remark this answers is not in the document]
MiriSafra: Intentional — `populate_vm_ids` is idempotent (same VM name → same inventory ID). All 3 DI test classes reuse the same config entry and just re-set the same `id` value. No test reads the config expecting `id` to be absent. The `prepared_plan` deepcopy in migration tests guards against non-idempotent mutations (storage/network map modifications) that don't apply to DI's read-only snapshot workflow.

utilities/deep_inspection.py
```diff
@@ -96,14 +105,120 @@ def create_di_connection_secret(
...
+    Args:
+        conversion (Conversion): The Conversion CR to cancel.
+    """
+    api_resource = conversion.client.resources.get(api_version=conversion.api_version, kind=conversion.kind)
+    api_resource.status.patch(
+        body={"status": {"phase": Conversion.Condition.CANCELED, "stage": CONVERSION_STAGE_FINISHED}},
+        name=conversion.name,
+        namespace=conversion.namespace,
+        content_type="application/merge-patch+json",
+    )
```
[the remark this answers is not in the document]
MiriSafra: No mismatch — `Conversion.Condition.CANCELED` resolves to the string `"Canceled"`, which is the exact phase string the forklift controller sets in `CancelConversion()` (`kubevirt.go:973-975`). There is no separate `Conversion.Status.CANCELED` constant in the SDK; `Condition.CANCELED` is the only available constant for this value. The `CONVERSION_TERMINAL_PHASES` set and `cancel_conversion()` both use the same resolved string.

myakove (commented):
## Code Review

Finding **1** (deepcopy mutation) — remaining findings are in the follow-up review below.

tests/deep_inspection/conftest.py
```diff
@@ -45,6 +47,24 @@ def di_connection_secret(
...
+    """First VM from the plan config with populated inventory ID.
+
+    Args:
+        class_plan_config (dict[str, Any]): Raw plan config from parametrization.
+        source_provider_inventory (ForkliftInventory): Source provider inventory.
+
+    Returns:
+        dict[str, Any]: VM dict with 'name' and 'id' keys.
+    """
+    populate_vm_ids(class_plan_config, source_provider_inventory)
```
myakove, line 86: ### [WARNING] di_resolved_vm mutates shared plan config in place

`class_plan_config` is the live `py_config["tests_params"][...]` object (via `request.param`). Calling `populate_vm_ids(class_plan_config, ...)` without a `deepcopy` permanently mutates that shared dict for every class that reuses `test_standalone_di_vsphere`.

The previous `di_conversion_resource` used `deepcopy(class_plan_config)` before populating IDs — restore that safeguard.

```suggestion
    plan_config = deepcopy(class_plan_config)
    populate_vm_ids(plan_config, source_provider_inventory)
    return plan_config["virtual_machines"][0]
```

(Also re-add `from copy import deepcopy`.)

utilities/deep_inspection.py
```diff
@@ -96,14 +105,120 @@ def create_di_connection_secret(
...
+    Mirrors the forklift controller's CancelConversion() pattern
+    (kubevirt.go:973-975): sets phase=Canceled, stage=Finished via
+    a status subresource merge-patch. The caller is responsible for
+    ensuring the conversion is in a cancelable state (e.g., Running).
+
+    Args:
+        conversion (Conversion): The Conversion CR to cancel.
+    """
+    api_resource = conversion.client.resources.get(api_version=conversion.api_version, kind=conversion.kind)
+    api_resource.status.patch(
```
myakove, line 122: ### [WARNING] Cancel via status subresource — confirm controller semantics

For `Migration` CRs, this codebase cancels via `spec.cancel` and waits for the controller to update status (`cancel_migration` in `utilities/migration_utils.py`). Here `cancel_conversion` patches `status.phase` / `status.stage` directly.

Status is normally controller-owned. If the Conversion reconciler overwrites an externally injected `phase=Canceled`, or only runs cleanup from an internal `CancelConversion()` path (not from watching status), then `test_verify_cancel_cleanup` / `test_verify_snapshot_cleanup` may pass or fail for reasons unrelated to real user-triggered cancel.

Please confirm against forklift that external status writes are the supported cancellation trigger for `Conversion`, and consider re-reading status after the patch to detect a reconciler race (phase reverting).

---
*Assisted-by: PI (cursor:grok-4.5[effort=high,fast=true])*

myakove (commented):
## Code Review

Found **11** issue(s) in this PR:

### :warning: Warnings (6)

| File | Line | Issue |
|------|------|-------|
| `utilities/deep_inspection.py` | 120 | Cancel via status subresource — confirm controller semantics |
| `tests/deep_inspection/test_standalone_di.py` | 131 | Cleanup polling belongs in utilities/, not test methods |
| `tests/deep_inspection/test_standalone_di.py` | 118 | Provider type-narrowing duplicated — validate at source |
| `utilities/deep_inspection.py` | 300 | Critical condition filter inconsistent with wait_for_critical_conditions |
| `tests/deep_inspection/test_standalone_di.py` | 161 | Dead assertion after get_vm_by_name |
| `tests/deep_inspection/test_standalone_di.py` | 91 | PR description does not match the implementation |

### :bulb: Suggestions (5)

| File | Line | Issue |
|------|------|-------|
| `tests/deep_inspection/test_standalone_di.py` | 81 | Unused plan config flag deep_inspection: True |
| `tests/deep_inspection/conftest.py` | 116 | Stale docstring still describes removed class_plan_config usage |
| `tests/deep_inspection/test_standalone_di.py` | 107 | No-op usefixtures(di_vm_name) |
| `tests/deep_inspection/test_standalone_di.py` | 120 | Pod-cleanup verification can false-positive if label selector is wrong |
| `tests/deep_inspection/conftest.py` | 62 | Returns wording implies only name/id keys |

---
*Review generated by pi*

---
*Assisted-by: PI (cursor:grok-4.5[effort=high,fast=true])*

tests/deep_inspection/test_standalone_di.py
```diff
@@ -49,3 +72,186 @@ def test_wait_for_completion(self) -> None:
...
+    @pytest.mark.usefixtures("di_vm_name")
+    def test_verify_cancel_cleanup(self) -> None:
+        """Verify canceled conversion reaches Canceled phase and pod is cleaned up."""
+        with pytest.raises(ConversionError, match="Canceled"):
+            wait_for_conversion_complete(
+                conversion=self.conversion,
+                timeout=py_config["plan_wait_timeout"],
+            )
+
+        try:
```
myakove: ### [WARNING] Cleanup polling belongs in utilities/, not test methods

`test_verify_cancel_cleanup` and `test_verify_snapshot_cleanup` embed substantial `TimeoutSampler` loops. Project file-placement rules put helper/polling logic in `utilities/` (same as `wait_for_di_snapshot`, `wait_for_conversion_phase`). Extract helpers such as `wait_for_conversion_pods_gone` / `wait_for_di_snapshot_removed` and keep the tests as thin orchestration.

---
*Assisted-by: PI (cursor:grok-4.5[effort=high,fast=true])*

tests/deep_inspection/test_standalone_di.py
```diff
@@ -49,3 +72,186 @@ def test_wait_for_completion(self) -> None:
...
+        """Wait for the Conversion CR to reach Running phase."""
+        wait_for_conversion_phase(
+            conversion=self.conversion,
+            phase=Conversion.Status.RUNNING,
+            timeout=py_config["plan_wait_timeout"],
+        )
+
+    def test_cancel_conversion(self, di_vm_name: str, source_provider: "BaseProvider") -> None:
+        """Wait for DI snapshot on source VM, then cancel the conversion."""
+        assert isinstance(source_provider, VMWareProvider), "Snapshot verification requires vSphere provider"
```
myakove: ### [WARNING] Provider type-narrowing duplicated — validate at source

The same `assert isinstance(source_provider, VMWareProvider)` appears here and in `test_verify_snapshot_cleanup`. Per Validate at Source / No Duplicate Code, move this into a class-scoped fixture (the PR body also claimed `vmware_source_provider`, which was never added) and inject that fixture instead of asserting in each test.

After adding a `vsphere_source_provider` / `vmware_source_provider` fixture that raises `TypeError` if not `VMWareProvider`, request that fixture here and drop the inline assert.

---
*Assisted-by: PI (cursor:grok-4.5[effort=high,fast=true])*

utilities/deep_inspection.py
```diff
@@ -177,20 +297,20 @@ def wait_for_conversion_complete(conversion: Conversion, timeout: int) -> None:
             if phase in CONVERSION_TERMINAL_PHASES:
                 if phase != Conversion.Status.SUCCEEDED:
                     conditions = sample.get("conditions", [])
-                    messages = [c.get("message", "") for c in conditions if c.get("type") == "Critical"]
+                    messages = [c.get("message", "") for c in conditions if c.get("category") == "Critical"]
```
myakove: ### [WARNING] Critical condition filter inconsistent with wait_for_critical_conditions

Here only `category == "Critical"` is checked. `wait_for_critical_conditions` also requires `status == "True"`. A Critical condition with `status: "False"` (cleared/inactive) would still appear in the `ConversionError` message and mislead diagnosis.

```suggestion
                    messages = [
                        c.get("message", "")
                        for c in conditions
                        if c.get("category") == "Critical" and c.get("status") == "True"
                    ]
```

---
*Assisted-by: PI (cursor:grok-4.5[effort=high,fast=true])*

tests/deep_inspection/test_standalone_di.py
```diff
@@ -49,3 +72,186 @@ def test_wait_for_completion(self) -> None:
...
+            )
+            raise AssertionError(
+                f"Conversion pods still present {DI_POD_CLEANUP_TIMEOUT}s after cancel: {[p.name for p in remaining]}"
+            ) from err
+
+    def test_verify_snapshot_cleanup(self, di_vm_name: str, source_provider: "BaseProvider") -> None:
+        """Verify forklift DI snapshot is removed from the source VM after cancel."""
+        assert isinstance(source_provider, VMWareProvider), "Snapshot verification requires vSphere provider"
+        vm = source_provider.get_vm_by_name(query=di_vm_name)
+        assert vm, f"VM '{di_vm_name}' not found in source provider"
```
myakove: ### [WARNING] Dead assertion after get_vm_by_name

`VMWareProvider.get_vm_by_name()` raises on a miss (`VmNotFoundError` / `ValueError`); it does not return a falsy value. This `assert vm, ...` is unreachable and implies a None-return contract that does not exist.

```suggestion
        vm = source_provider.get_vm_by_name(query=di_vm_name)
```

---
*Assisted-by: PI (cursor:grok-4.5[effort=high,fast=true])*

tests/deep_inspection/test_standalone_di.py
```diff
@@ -49,3 +72,186 @@ def test_wait_for_completion(self) -> None:
...
+        ),
+    ],
+    indirect=True,
+    ids=["standalone-di-cancel-vsphere"],
+)
+@pytest.mark.incremental
+@pytest.mark.tier1
+@pytest.mark.deep_inspection
+@pytest.mark.vsphere
+class TestStandaloneDICancel:
```
myakove, line 88: ### [WARNING] PR description does not match the implementation

- Claims new fixtures `vmware_source_provider` and `di_rerun_conversion_resource` — neither exists. Rerun builds the Conversion inline in `test_rerun_di_after_cancel` via `create_conversion_resource(...)`.
- Says `TestStandaloneDICancel` has 6 tests; the class defines 8 methods (`test_create_conversion` through `test_rerun_verify_results`).

Please update the PR body so reviewers can trust the deliverables list.

---
*Assisted-by: PI (cursor:grok-4.5[effort=high,fast=true])*

tests/deep_inspection/test_standalone_di.py
```diff
@@ -49,3 +72,186 @@ def test_wait_for_completion(self) -> None:
     def test_verify_di_results(self, di_vm_name: str) -> None:
         """Verify inspection results contain OS info and filesystem data."""
         verify_di_results(conversion=self.conversion, vm_name=di_vm_name)
+
+
+@pytest.mark.parametrize(
+    "class_plan_config",
+    [
+        pytest.param(
+            py_config["tests_params"]["test_standalone_di_vsphere"],
```
myakove, line 78: ### [SUGGESTION] Unused plan config flag deep_inspection: True

`tests/tests_config/config.py` sets `"deep_inspection": True` on `test_standalone_di_vsphere`, but nothing in the codebase reads `plan.get("deep_inspection")` (or similar). Gating uses `@pytest.mark.deep_inspection` instead.

If the key is unused, remove it from config to avoid implying it drives behavior. If it was meant to gate something, that wiring appears missing.

---
*Assisted-by: PI (cursor:grok-4.5[effort=high,fast=true])*

tests/deep_inspection/conftest.py
```diff
@@ -75,34 +113,56 @@ def di_conversion_resource(
     is needed.
 
     Args:
-        class_plan_config (dict[str, Any]): Raw plan config from parametrization.
+        di_resolved_vm (dict[str, Any]): Resolved VM dict with 'name' and 'id'.
```
myakove: ### [SUGGESTION] Stale docstring still describes removed class_plan_config usage

The fixture prose above still says it uses `class_plan_config` directly, but the fixture now takes `di_resolved_vm` / `di_vddk_image`. Update that paragraph to match the current Args list (e.g. "Uses di_resolved_vm (not prepared_plan)...").

---
*Assisted-by: PI (cursor:grok-4.5[effort=high,fast=true])*

tests/deep_inspection/test_standalone_di.py
```diff
@@ -49,3 +72,186 @@ def test_wait_for_completion(self) -> None:
...
+
+    conversion: Conversion
+    rerun_conversion: Conversion
+
+    def test_create_conversion(self, di_conversion_resource: Conversion) -> None:
+        """Create a standalone DeepInspection Conversion CR for cancel testing."""
+        self.__class__.conversion = di_conversion_resource
+        assert self.conversion
+
+    @pytest.mark.usefixtures("di_vm_name")
```
myakove: ### [SUGGESTION] No-op usefixtures(di_vm_name)

`di_vm_name` is a pure value fixture with no side effects. Requesting it via `usefixtures` on methods that never use the name (`test_wait_for_running`, `test_verify_cancel_cleanup`, `test_rerun_wait_for_completion`, `test_verify_validation_error`) does nothing. Remove those markers (Fixture Request Patterns: only request fixtures that are needed).

---
*Assisted-by: PI (cursor:grok-4.5[effort=high,fast=true])*

tests/deep_inspection/test_standalone_di.py
```diff
@@ -49,3 +72,186 @@ def test_wait_for_completion(self) -> None:
...
+            conversion=self.conversion,
+            phase=Conversion.Status.RUNNING,
+            timeout=py_config["plan_wait_timeout"],
+        )
+
+    def test_cancel_conversion(self, di_vm_name: str, source_provider: "BaseProvider") -> None:
+        """Wait for DI snapshot on source VM, then cancel the conversion."""
+        assert isinstance(source_provider, VMWareProvider), "Snapshot verification requires vSphere provider"
+        wait_for_di_snapshot(source_provider=source_provider, vm_name=di_vm_name)
+        cancel_conversion(conversion=self.conversion)
```
myakove, line 117: ### [SUGGESTION] Pod-cleanup verification can false-positive if label selector is wrong

`test_verify_cancel_cleanup` treats an empty pod list as success. If `conversion=<name>` is not the label the DI pod actually carries, `Pod.get()` returns `[]` on the first sample and the test passes without ever observing a real pod.

Assert that at least one matching pod exists **before** calling `cancel_conversion`:

```suggestion
        wait_for_di_snapshot(source_provider=source_provider, vm_name=di_vm_name)
        pods_before_cancel = list(
            Pod.get(
                client=self.conversion.client,
                namespace=self.conversion.namespace,
                label_selector=f"conversion={self.conversion.name}",
            )
        )
        assert pods_before_cancel, (
            f"No pods found for conversion '{self.conversion.name}' before cancel — "
            "label selector may not match the running conversion pod"
        )
        cancel_conversion(conversion=self.conversion)
```

---
*Assisted-by: PI (cursor:grok-4.5[effort=high,fast=true])*

tests/deep_inspection/conftest.py
```diff
@@ -45,6 +47,24 @@ def di_connection_secret(
...
+    source_provider_inventory: "ForkliftInventory",
+) -> dict[str, Any]:
+    """First VM from the plan config with populated inventory ID.
+
+    Args:
+        class_plan_config (dict[str, Any]): Raw plan config from parametrization.
+        source_provider_inventory (ForkliftInventory): Source provider inventory.
+
+    Returns:
+        dict[str, Any]: VM dict with 'name' and 'id' keys.
```
myakove: ### [SUGGESTION] Returns wording implies only name/id keys

The fixture returns `class_plan_config["virtual_machines"][0]` after populating `id`. That dict can still carry other plan keys (`source_vm_power`, `guest_agent`, etc.). Prefer wording like: "VM config dict, guaranteed to include `'name'` and `'id'`."

---
*Assisted-by: PI (cursor:grok-4.5[effort=high,fast=true])*

tests/deep_inspection/conftest.py
MiriSafra: Intentionally not adding deepcopy. krcmarik's prior review (comment #3573724897) explicitly asked to remove deepcopy from this fixture, and `populate_vm_ids` is idempotent — same VM name always resolves to the same inventory ID. Adding deepcopy here would contradict the prior review decision.

utilities/deep_inspection.py
MiriSafra: Good question. The cancel mechanism for Conversion CRs is different from Migration CRs by design. Forklift's `CancelConversion()` (kubevirt.go:973-975) patches the status subresource directly with `phase=Canceled, stage=Finished`. The Conversion reconciler checks `status.phase == Canceled` on each reconcile loop and runs cleanup (delete pods, remove snapshot). This is the documented controller pattern — not a spec field like Migration's `spec.cancel`. Tested and working on cluster.

tests/deep_inspection/test_standalone_di.py
MiriSafra: Done. Extracted `wait_for_conversion_pods_cleanup()` and `wait_for_di_snapshot_cleanup()` into `utilities/deep_inspection.py`. Test methods now call these utilities.

tests/deep_inspection/test_standalone_di.py
MiriSafra: Done. Added `vmware_source_provider` fixture in conftest.py that validates `isinstance` and raises `TypeError`. Test methods now use this fixture instead of inline isinstance asserts.

utilities/deep_inspection.py
MiriSafra: Done. Added `and c.get("status") == "True"` to the condition filter in `wait_for_conversion_complete`, consistent with `wait_for_critical_conditions`.

tests/deep_inspection/test_standalone_di.py
MiriSafra: Done. Removed dead `assert vm` — `get_vm_by_name` raises on miss, never returns None.

tests/deep_inspection/test_standalone_di.py
MiriSafra: Done. Updated PR description: fixed test count (8 tests), removed non-existent `di_rerun_conversion_resource` fixture reference, added new utility functions.

tests/deep_inspection/test_standalone_di.py
MiriSafra: Done. Removed `"deep_inspection": True` — nothing reads this key.

tests/deep_inspection/conftest.py
MiriSafra: Done. Updated docstring to remove `class_plan_config` reference. Now says "Bypasses prepared_plan because DI only creates a snapshot".

tests/deep_inspection/test_standalone_di.py
MiriSafra: Good catch that `di_vm_name` is a pure value. However, the decorator can't be fully removed — it maintains the fixture dependency chain needed for `indirect=True` parametrization. Without it, pytest errors: "function uses no fixture 'class_plan_config'". Changed to `@pytest.mark.usefixtures("class_plan_config")` which makes the actual dependency explicit.

tests/deep_inspection/test_standalone_di.py
MiriSafra: Done. Added pod existence assertion before `cancel_conversion()` in `test_cancel_conversion`. Uses `Pod.get()` with `label_selector=f"conversion={self.conversion.name}"` and asserts non-empty.

tests/deep_inspection/conftest.py
MiriSafra: Done. Updated to "VM config dict with populated 'name' and 'id' keys".

rh-bot-1:
**Clean rebase detected** — no code changes compared to previous head (`48f2bff`).
The following labels were preserved: `lgtm-coderabbitai[bot]`, `commented-MiriSafra`.

utilities/deep_inspection.py
```diff
@@ -15,10 +16,19 @@
...
 
 CONVERSION_TERMINAL_PHASES = {Conversion.Status.SUCCEEDED, Conversion.Status.FAILED, Conversion.Condition.CANCELED}
+# Not exposed by openshift-python-wrapper — mirrors forklift constants:
+CONVERSION_TYPE_DEEP_INSPECTION = "DeepInspection"  # forklift: DeepInspection (conversion.go:20)
+CONVERSION_STAGE_FINISHED = "Finished"  # forklift: StageFinished (conversion.go:55)
+VDDK_IMAGE_NOT_SET_CONDITION = "VDDKImageNotSet"  # forklift: validation.go:14
+DI_SNAPSHOT_NAME = "forklift-deep-inspection"  # forklift: snapshotName (client.go:21)
+DI_SNAPSHOT_CREATION_TIMEOUT = 120
+DI_POD_CLEANUP_TIMEOUT = 60
+DI_SNAPSHOT_CLEANUP_TIMEOUT = 180
```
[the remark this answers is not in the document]
MiriSafra: Skipped: AI nitpick — current placement matches repo convention

utilities/deep_inspection.py
```diff
@@ -230,3 +355,175 @@ def verify_di_results(conversion: Conversion, vm_name: str) -> None:
...
+        raise AssertionError(
+            f"Conversion '{conversion.name}' expected Critical conditions within {timeout}s. "
+            f"Last phase='{phase}', conditions={last_condition_types}"
+        ) from err
+
+    return critical, phase
+
+
+DI_POD_CREATION_TIMEOUT = 120
+
```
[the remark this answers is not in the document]
MiriSafra: Fixed in f2ecb97: moved `DI_POD_CREATION_TIMEOUT` to the top-of-file constants block alongside `DI_SNAPSHOT_CREATION_TIMEOUT` and `DI_POD_CLEANUP_TIMEOUT`.

tests/deep_inspection/test_standalone_di.py
```diff
@@ -49,3 +69,146 @@ def test_wait_for_completion(self) -> None:
...
+            conversion=self.conversion,
+            phase=Conversion.Status.RUNNING,
+            timeout=py_config["plan_wait_timeout"],
+        )
+
+    def test_cancel_conversion(self, di_vm_name: str, vmware_source_provider: "VMWareProvider") -> None:
+        """Wait for DI pod to be running on the cluster, then cancel the conversion."""
+        wait_for_di_snapshot(source_provider=vmware_source_provider, vm_name=di_vm_name)
+        wait_for_conversion_pods(conversion=self.conversion)
+        cancel_conversion(conversion=self.conversion)
```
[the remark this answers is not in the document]
MiriSafra: Intentionally keeping 120s default. This timeout is for pod creation (controller API call to create the pod after snapshot), not for the full DI pipeline (3600s). The controller creates the snapshot first, then issues a single API call to create the pod — 120s is generous for that. Using 3600s would mask real pod-creation failures by waiting an hour before surfacing them. The terminal phase check added in f2ecb97 (#3 below) provides the early-exit safety net for cases where the conversion fails before pod creation.

utilities/deep_inspection.py
```diff
@@ -230,3 +355,175 @@ def verify_di_results(conversion: Conversion, vm_name: str) -> None:
...
+                    namespace=conversion.namespace,
+                    label_selector=f"conversion={conversion.name}",
+                )
+            ),
+        ):
+            if pods:
+                LOGGER.info(f"Found {len(pods)} pod(s) for conversion '{conversion.name}'")
+                return pods
+    except TimeoutExpiredError as err:
+        raise AssertionError(f"No conversion pods found for '{conversion.name}' within {timeout}s") from err
```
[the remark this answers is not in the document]
MiriSafra: Fixed in f2ecb97: `wait_for_conversion_pods()` now checks `conversion.instance.status.phase` on each poll iteration. If the phase is in `CONVERSION_TERMINAL_PHASES` (Failed/Canceled) before any pod appears, it raises `ConversionError` immediately with the phase info instead of waiting until timeout.

utilities/deep_inspection.py
```diff
@@ -230,3 +356,189 @@ def verify_di_results(conversion: Conversion, vm_name: str) -> None:
...
+        list[Pod]: The conversion pods found.
+
+    Raises:
+        ConversionError: If the conversion reaches a terminal phase before pods appear.
+        AssertionError: If no pods appear within timeout.
+    """
+
+    def _poll() -> list["Pod"]:
+        status = conversion.instance.status or {}
+        phase = status.get("phase", "")
```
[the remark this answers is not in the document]
MiriSafra: Fixed in 27c9c34: wrapped `conversion.instance.status` access in `try/except NotFoundError`, re-raising as `ConversionError` with context and exception chaining. Matches the existing `NotFoundError` handling pattern at lines 64 and 79. Updated docstring Raises section.

PR #593, opened 2026-07-09
fix(MTV-6066): stabilize vSphere Forklift inventory sync during prepared_plan

prabinovRedhat:
## Summary

Fixes MTV-6066 vSphere Forklift inventory sync failures during `prepared_plan` by introducing a two-phase clone workflow and targeted inventory wait/refresh workarounds.

## Problem

Cloned VMs were not reliably appearing in Forklift inventory before migration planning, causing intermittent failures in copy-offload and other vSphere tests during `prepared_plan`.

## Changes

- **Two-phase clone workflow**: clone all VMs first, then batch inventory sync via `wait_for_cloned_vms_in_forklift_inventory()`
- **`wait_for_hosts()` / `wait_for_datastores()`**: wait for host and datastore inventory before VM sync
- **Force inventory refresh on timeout**: call `force_inventory_refresh()` when VM inventory sync times out
- **MTV-6072 gating**: all three workarounds are gated on `jira_issue_open("MTV-6072")` via new `utilities/jira_helpers.py` so they auto-disable when the upstream bug is resolved
- **Refactors**: extract copy-offload datastore helpers, improve VMware provider inventory handling, update AGENTS.md

## Files changed

- `conftest.py` — two-phase clone in `prepared_plan`
- `libs/forklift_inventory.py` — inventory wait helpers
- `utilities/provider_inventory.py` — `wait_for_cloned_vms_in_forklift_inventory()` orchestration
- `utilities/copyoffload_datastore.py` — extracted datastore utilities
- `utilities/jira_helpers.py` — Jira runtime issue state helper
- `libs/providers/vmware.py` — inventory refresh support
- `tests/copyoffload/test_copyoffload_migration.py` — test adjustment
- `AGENTS.md` — document MTV-6072 gating pattern

## Test plan

- [x] Jenkins build 396 passed `TestCopyoffloadMultiDatastoreMigration` (MTV-564) on this branch
- [ ] Run copy-offload tests with cloned VMs on vSphere provider
- [ ] Verify workarounds are active when MTV-6072 is open and skipped when resolved

## Commits

1. `2827aa6` fix(MTV-6066): wait for hosts and two-phase VM inventory sync
2. `905bd11` fix(MTV-6066): wait for datastores before VM inventory sync
3. `be00262` fix(MTV-6066): force provider inventory refresh on VM sync timeout
4. `a2d6acc` refactor(MTV-6066): gate inventory workarounds on MTV-6072 Jira state
5. `fd597f2` refactor(MTV-6066): address review fixes for inventory sync

Fixes MTV-6066

Made with [Cursor](https://cursor.com)

## Summary by CodeRabbit

* **New Features**
  * Enhanced multi-datastore VM cloning with improved datastore identifier resolution and validation.
  * Added vSphere inventory synchronization improvements, including host/datastore availability checks and a two-phase cloned-VM inventory wait workflow.
  * Introduced an optional per-VM inventory wait timeout (`inventory_timeout`, default 300s), with Jira-gated workaround behavior.

* **Bug Fixes**
  * Reduced inventory synchronization failures for multi-VM, cross-datastore clones.
  * Added clearer error messaging when datastores are missing, empty, or unavailable.

* **Tests**
  * Updated/adjusted copyoffload migration test coverage for new configuration expectations.


utilities/provider_inventory.py
```diff
@@ -0,0 +1,201 @@
...
+                _retry_vsphere_vm_inventory_after_refresh(
+                    source_provider=source_provider,
+                    source_provider_inventory=vsphere_inventory,
+                    vm_name=vm_name,
+                    virtual_machines=virtual_machines,
+                    copyoffload_config=copyoffload_config,
+                    inventory_timeout=inventory_timeout,
+                )
+            else:
+                raise
```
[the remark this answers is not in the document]
prabinovRedhat: Fixed in fd2fcc0 — refactored to collect failed VMs, one shared refresh+rewait, retry failed subset only.

utilities/provider_inventory.py
```diff
@@ -0,0 +1,201 @@
...
+        TimeoutExpiredError: If the VM does not appear in inventory within the timeout
+    """
+    force_inventory_refresh(source_provider.ocp_resource)
+    _wait_for_vsphere_host_and_datastore_inventory(
+        source_provider_inventory=source_provider_inventory,
+        virtual_machines=virtual_machines,
+        copyoffload_config=copyoffload_config,
+        inventory_timeout=inventory_timeout,
+    )
+    source_provider_inventory.wait_for_vm(name=vm_name, timeout=inventory_timeout)
```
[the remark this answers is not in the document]
prabinovRedhat: Fixed in fd2fcc0 — explicit ValueError before force_inventory_refresh.

utilities/copyoffload_datastore.py
```diff
@@ -0,0 +1,42 @@
+from typing import Any
```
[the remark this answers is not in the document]
prabinovRedhat: Fixed in fd2fcc0 — added `_configured_datastore_moid()` with str validation.

tests/copyoffload/test_copyoffload_migration.py
```diff
@@ -2013,6 +2013,7 @@ def test_check_vms(
     "vmware_cloud_init_ready",
     "multus_network_name",
     "copyoffload_config",
+    "multi_datastore_config",
```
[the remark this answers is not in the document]
prabinovRedhat: Fixed in fd2fcc0 — removed redundant guard; multi_datastore_config fixture owns validation.

utilities/jira_helpers.py
```diff
@@ -0,0 +1,28 @@
...
+    if jira_plugin:
+        try:
+            return not jira_plugin.is_issue_resolved(issue_id)
+        except requests.RequestException as e:
+            strategy = request.config.getoption(CONNECTION_ERROR_FLAG_NAME)
+            if strategy == SKIP:
+                pytest.skip(CONNECTION_SKIP_MESSAGE % e)
+            elif strategy == STRICT:
+                raise
+    return None
```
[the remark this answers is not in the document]
prabinovRedhat: Fixed in fd2fcc0 — added Raises section for requests.RequestException.

libs/providers/vmware.py
```diff
@@ -1051,27 +1046,29 @@ def _resolve_datastore(disk_datastore_id: str | None) -> vim.Datastore:
...
-                    raise VmCloneError(
-                        "Disk requested non-XCOPY datastore but copyoffload.non_xcopy_datastore_id is not configured"
-                    )
-                resolved = non_xcopy_datastore
-            elif disk_datastore_id is not None:
+            if disk_datastore_id is None:
+                resolved = target_datastore
+            else:
                 if not disk_datastore_id:
                     raise VmCloneError("Disk datastore_id is empty. Provide a valid MoID or omit the field.")
```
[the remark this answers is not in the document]
prabinovRedhat: Fixed in fd2fcc0 — moved messages to constants/formatters in copyoffload_datastore.py.

libs/forklift_inventory.py
```diff
@@ -406,10 +406,106 @@ def __init__(self, client: DynamicClient, provider_name: str, namespace: str) ->
...
+            f"Waiting for datastores {sorted(requested_ids)} to appear in Forklift inventory "
+            f"for provider '{self.provider_name}'..."
+        )
+
+        def _check_datastores() -> list[dict[str, Any]] | None:
+            storages = self.storages
+            found_ids = {storage["id"] for storage in storages if storage.get("id")}
+            if requested_ids - found_ids:
+                return None
+            return [storage for storage in storages if storage.get("id") in requested_ids]
```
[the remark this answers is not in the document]
prabinovRedhat: Fixed in fd2fcc0 — extracted `_extract_storage_ids()`.

libs/forklift_inventory.py
```diff
@@ -406,10 +406,106 @@ def __init__(self, client: DynamicClient, provider_name: str, namespace: str) ->
...
+                sleep=sleep,
+                func=lambda: self.hosts,
+            ):
+                if sample:
+                    LOGGER.info(f"Found {len(sample)} hosts in inventory for provider '{self.provider_name}'")
+                    return sample
+        except TimeoutExpiredError:
+            raise TimeoutExpiredError(
+                f"No hosts appeared in Forklift inventory for provider '{self.provider_name}' after {timeout}s"
+            )
```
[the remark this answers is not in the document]
prabinovRedhat: Fixed in fd2fcc0 — added `from None` for consistency with wait_for_datastores.

prabinovRedhat:
Addressed all CodeRabbit review feedback in fd2fcc0 (`fix(MTV-6066): address CodeRabbit PR review feedback`):

- **forklift_inventory** — added `from None` on `wait_for_hosts` re-raise for consistency with `wait_for_datastores`
- **forklift_inventory** — extracted `_extract_storage_ids()` to deduplicate storage-ID extraction
- **vmware.py** — moved inline `VmCloneError` messages to constants/formatters in `copyoffload_datastore.py` (TRY003)
- **test_copyoffload_migration.py** — removed redundant `pytest.fail` guard; `multi_datastore_config` fixture owns validation
- **copyoffload_datastore.py** — added `_configured_datastore_moid()` with str validation for symbolic datastore config
- **jira_helpers.py** — added `Raises` section documenting `requests.RequestException`
- **provider_inventory.py** — explicit `ValueError` before `force_inventory_refresh` when `ocp_resource` is missing
- **provider_inventory.py** — batch refresh refactor: collect failed VMs, one shared refresh+rewait, retry failed subset only

Replies posted on each inline thread.

prabinovRedhat:
Screenshot 2026-07-09 at 23 51 44

utilities/jira_helpers.py
```diff
@@ -0,0 +1,31 @@
...
```
myakove: I whould add here logger warning if no jira plugin for visabilty.

utilities/provider_inventory.py
```diff
@@ -0,0 +1,195 @@
...
+            )
+        vsphere_inventory = source_provider_inventory
+        _wait_for_vsphere_host_and_datastore_inventory(
+            source_provider_inventory=vsphere_inventory,
+            virtual_machines=virtual_machines,
+            copyoffload_config=copyoffload_config,
+            inventory_timeout=inventory_timeout,
+        )
+
+    failed_vm_names: list[str] = []
```
myakove: **Duplicated wait-for-VM loop**

The "loop over VM names, try `wait_for_vm`, collect failures" pattern is identical at lines 156-161 and 180-185. Per "No Duplicate Code (MUST)", extract a helper.

This also brings `wait_for_cloned_vms_in_forklift_inventory` under the 50-line guideline.

**Suggested helper:**
```python
def _collect_inventory_wait_failures(
    source_provider_inventory: ForkliftInventory,
    vm_names: list[str],
    timeout: int,
) -> list[str]:
    failed: list[str] = []
    for vm_name in vm_names:
        try:
            source_provider_inventory.wait_for_vm(name=vm_name, timeout=timeout)
        except TimeoutExpiredError:
            failed.append(vm_name)
    return failed
```
Then replace both loops with calls to this helper.

---

utilities/provider_inventory.py
```diff
@@ -0,0 +1,195 @@
...
+        try:
+            source_provider_inventory.wait_for_vm(name=vm_name, timeout=inventory_timeout)
+        except TimeoutExpiredError:
+            failed_vm_names.append(vm_name)
+
+    if not failed_vm_names:
+        return
+
+    if is_vsphere and workaround_active and vsphere_inventory is not None:
+        LOGGER.info(
```
myakove: **Use f-string for non-expensive logging**

`failed_vm_names` is a plain list of strings — not an expensive operation. Per logging rules, use f-strings by default; `%s` is only for expensive operations where lazy evaluation matters.

```suggestion
        LOGGER.info(
            f"VM inventory wait timed out for {failed_vm_names}; forcing provider refresh and retrying"
        )
```

---

myakove (commented):
## Code Review

### :bulb: Suggestions (6)

| File | Line | Issue |
|------|------|-------|
| `utilities/provider_inventory.py` | 156 | **Duplicated wait-for-VM loop** |
| `utilities/provider_inventory.py` | 167 | **Use f-string for non-expensive logging** |
| `utilities/copyoffload_datastore.py` | 42 | **Use existing constants instead of hardcoded string literals** |
| `libs/forklift_inventory.py` | 422 | **Missing docstring on new `hosts` property** |
| `utilities/jira_helpers.py` | 28 | **`pytest.skip()` in utility function violates No Auto-Skip rule** |
| `utilities/provider_inventory.py` | 155 | **Consider: cumulative timeout can be N × timeout** |

---
*Review generated by pi*

---

utilities/copyoffload_datastore.py
```diff
@@ -0,0 +1,74 @@
...
+
+    Returns:
+        Resolved vSphere datastore MoID
+
+    Raises:
+        ValueError: If the key is missing or the value is not a non-empty string
+    """
+    resolved_id = copyoffload_config.get(key)
+    if not resolved_id:
+        if key == "secondary_datastore_id":
```
myakove: **Use existing constants instead of hardcoded string literals**

`_configured_datastore_moid` uses `"secondary_datastore_id"` / `"non_xcopy_datastore_id"` string literals, while `SYMBOLIC_SECONDARY_DATASTORE` / `SYMBOLIC_NON_XCOPY_DATASTORE` constants exist for these exact values. Same issue in `resolve_datastore_moid_from_disk_config` at line 68. Use constants consistently to avoid drift.

```suggestion
        if key == SYMBOLIC_SECONDARY_DATASTORE:
            raise ValueError(ERR_SECONDARY_DS_NOT_CONFIGURED)
        if key == SYMBOLIC_NON_XCOPY_DATASTORE:
            raise ValueError(ERR_NON_XCOPY_DS_NOT_CONFIGURED)
```

And in `resolve_datastore_moid_from_disk_config`:
```python
    if disk_datastore_id == SYMBOLIC_SECONDARY_DATASTORE:
        return _configured_datastore_moid(copyoffload_config, SYMBOLIC_SECONDARY_DATASTORE)

    if disk_datastore_id == SYMBOLIC_NON_XCOPY_DATASTORE:
        return _configured_datastore_moid(copyoffload_config, SYMBOLIC_NON_XCOPY_DATASTORE)
```

---

libs/forklift_inventory.py
```diff
@@ -406,10 +418,106 @@ def __init__(self, client: DynamicClient, provider_name: str, namespace: str) ->
             client=client, provider_name=provider_name, mtv_namespace=namespace, provider_type=self.provider_type
         )
 
+    @property
+    def hosts(self) -> list[dict[str, Any]]:
```
myakove, line 422: **Missing docstring on new `hosts` property**

Per "Docstring Format (MUST)": all new functions must have docstrings.

```suggestion
    @property
    def hosts(self) -> list[dict[str, Any]]:
        """List ESXi hosts from the Forklift vSphere provider inventory."""
        return self._request(url_path=f"{self.provider_url_path}/hosts")
```

---

utilities/jira_helpers.py
```diff
@@ -0,0 +1,31 @@
...
+        requests.RequestException: If Jira connection fails and connection strategy is STRICT
+    """
+    jira_plugin = request.config.pluginmanager.getplugin(JIRA_PLUGIN_NAME)
+    if jira_plugin:
+        try:
+            return not jira_plugin.is_issue_resolved(issue_id)
+        except requests.RequestException as e:
+            strategy = request.config.getoption(CONNECTION_ERROR_FLAG_NAME)
+            if strategy == SKIP:
+                pytest.skip(CONNECTION_SKIP_MESSAGE % e)
```
myakove: **`pytest.skip()` in utility function violates No Auto-Skip rule**

When Jira is unreachable and strategy is `SKIP`, `pytest.skip()` silently skips the entire test class from inside a utility function called during fixture setup. This violates the "No Auto-Skip" rule and breaks the documented `-> bool | None` return contract — the function never returns when strategy is SKIP.

Should return `None` + log warning instead, letting the caller decide:

```suggestion
            if strategy == SKIP:
                LOGGER.warning(f"Jira connection failed (strategy=SKIP): {e}")
            elif strategy == STRICT:
                raise
```

---

utilities/provider_inventory.py
```diff
@@ -0,0 +1,195 @@
...
+                f"vSphere provider requires VsphereForkliftInventory, got {type(source_provider_inventory).__name__}"
+            )
+        vsphere_inventory = source_provider_inventory
+        _wait_for_vsphere_host_and_datastore_inventory(
+            source_provider_inventory=vsphere_inventory,
+            virtual_machines=virtual_machines,
+            copyoffload_config=copyoffload_config,
+            inventory_timeout=inventory_timeout,
+        )
+
```
myakove: **Consider: cumulative timeout can be N × timeout**

Each failed VM consumes the full `inventory_timeout` before the next is checked. With 5 VMs and a 300s timeout, worst case is 1500s (25 min) for the first attempt alone, plus another 1500s for retry.

Consider documenting this behavior in the docstring, or adding a total elapsed-time guard to cap overall wait time.

---

krcmarik:
Honestly I am not really sure If we should workaround what I believe is either bug in forklift or the vmware environment misconfiguration in the test suite. We are supposed to get almost immediate updates from VMware about any inventory changes. If we do not have them within 300s, something is wrong and forcing inventory refresh in the test suite has two consequences:
1. We are imo covering bugs of the tested product in the test suite which goes against the entire point of test automation.
2. If the problem keeps happening and It seems it does we extend the test run time by 300s with each failure of clone VM's discovery in the forklift inventory.

myakove:
> Honestly I am not really sure If we should workaround what I believe is either bug in forklift or the vmware environment misconfiguration in the test suite. We are supposed to get almost immediate updates from VMware about any inventory changes. If we do not have them within 300s, something is wrong and forcing inventory refresh in the test suite has two consequences:
> 1. We are imo covering bugs of the tested product in the test suite which goes against the entire point of test automation.
> 2. If the problem keeps happening and It seems it does we extend the test run time by 300s with each failure of clone VM's discovery in the forklift inventory.

You are right, but we have 2 options.
Fail the test on the inventory bug or handle it in the test for know till the bug is fixed (the wa taid to Jira ticket)
The Jira way it the right way to handle it, while the bug is open we workaround it, once it's closed we go back to the original code flow without the workaround 

I think this option is better for us, we do not want to not test the migration completely because the inventory bug.

myakove:
For your second concerned (test time) we need to try to change the code and try to find a better way to handle it if possible.

utilities/provider_inventory.py
```diff
@@ -0,0 +1,195 @@
...
+    if not failed_vm_names:
+        return
+
+    if is_vsphere and workaround_active and vsphere_inventory is not None:
+        LOGGER.info(
+            "VM inventory wait timed out for %s; forcing provider refresh and retrying",
+            failed_vm_names,
+        )
+        if source_provider.ocp_resource is None:
```
krcmarik: This check runs after the initial host/datastore wait + per-VM wait have already consumed their full timeout budget. If `ocp_resource` is None and we know `is_vsphere and workaround_active` at the top of the function, we already know the retry path will need it.

Consider validating `source_provider.ocp_resource` upfront (right after `if is_vsphere and workaround_active:`) so a config issue fails immediately rather than after potentially minutes of waiting.

utilities/provider_inventory.py
```diff
@@ -0,0 +1,195 @@
...
+        provider (Provider): Forklift Provider resource to refresh.
+
+    Raises:
+        TimeoutExpiredError: If the provider does not become Ready within the timeout.
+    """
+    refresh_timestamp = str(int(time.time()))
+    LOGGER.info(f"Forcing inventory refresh for provider '{provider.name}' with _refresh={refresh_timestamp}")
+    patch = {"spec": {"settings": {"_refresh": refresh_timestamp}}}
+    ResourceEditor(patches={provider: patch}).update()
+    provider.wait_for_condition(condition="Ready", status="True", timeout=_INVENTORY_REFRESH_READY_TIMEOUT)
```
krcmarik, line 39: The existing VMware provider code (`libs/providers/vmware.py`, `_wait_for_vsphere_provider_ready`) waits for `Validated` after patching `spec.settings`. This waits for `Ready` instead. `Ready=True` can occur before `Validated=True` — if the provider has validation errors, inventory queries after this point could fail with misleading errors rather than a clear "provider not validated" message.

Is `Ready` intentionally sufficient here, or should this match the existing `Validated` pattern?

utilities/copyoffload_datastore.py
```diff
@@ -0,0 +1,74 @@
...
+            raise ValueError(ERR_SECONDARY_DS_NOT_CONFIGURED)
+        if key == "non_xcopy_datastore_id":
+            raise ValueError(ERR_NON_XCOPY_DS_NOT_CONFIGURED)
+        raise ValueError(f"copyoffload.{key} is not configured")
```
krcmarik, line 47: This line is unreachable — `_configured_datastore_moid` is only called from `resolve_datastore_moid_from_disk_config` with exactly `"secondary_datastore_id"` or `"non_xcopy_datastore_id"`, both handled by the preceding if-branches.

libs/providers/vmware.py
```diff
@@ -1051,27 +1051,27 @@ def _resolve_datastore(disk_datastore_id: str | None) -> vim.Datastore:
...
+                    raise VmCloneError(ERR_EMPTY_DISK_DATASTORE_ID)
                 try:
-                    resolved = self.get_obj([vim.Datastore], disk_datastore_id)
-                except ValueError:
-                    raise VmCloneError(
-                        f"Custom datastore not found for disk. MoID '{disk_datastore_id}' is invalid or not accessible."
-                    ) from None
-            else:
-                resolved = target_datastore
+                    resolved_moid = resolve_datastore_moid_from_disk_config(
```
krcmarik: This round-trip resolves the symbolic key to a MoID via `resolve_datastore_moid_from_disk_config`, then compares the MoID against `secondary_datastore._moId` / `non_xcopy_datastore._moId` — which are guaranteed to match since they were resolved from the same config key earlier in the function.

The old code simply checked `disk_datastore_id == "secondary_datastore_id"` and used the pre-resolved object directly. The refactored version is functionally correct but adds indirection without behavioral gain for the known symbolic keys.

libs/forklift_inventory.py
```diff
@@ -406,10 +418,106 @@ def __init__(self, client: DynamicClient, provider_name: str, namespace: str) ->
...
+                if sample:
+                    LOGGER.info(f"Found {len(sample)} hosts in inventory for provider '{self.provider_name}'")
+                    return sample
+        except TimeoutExpiredError:
+            raise TimeoutExpiredError(
+                f"No hosts appeared in Forklift inventory for provider '{self.provider_name}' after {timeout}s"
+            ) from None
+
+        # This should never be reached, but satisfies type checker
+        raise TimeoutExpiredError("Host wait completed unexpectedly without returning")
```
krcmarik: This line is unreachable — `TimeoutSampler` either yields a truthy value (returned at the `if sample:` check) or raises `TimeoutExpiredError` (caught by the except block). Same applies to the identical pattern at the end of `wait_for_datastores`.

utilities/jira_helpers.py
```diff
@@ -0,0 +1,31 @@
+import pytest
+import requests
+from pytest_jira import CONNECTION_ERROR_FLAG_NAME, CONNECTION_SKIP_MESSAGE, SKIP, STRICT
+
+JIRA_PLUGIN_NAME = "jira_plugin"
+
+
+def is_jira_issue_open(request: pytest.FixtureRequest, issue_id: str) -> bool | None:
```
krcmarik: This function reimplements the built-in `jira_issue` fixture from pytest-jira (`pytest_jira.py:640-664`). The logic is nearly identical: get plugin → call `is_issue_resolved` → invert result → handle `RequestException` with SKIP/STRICT strategy.

The only reason for this reimplementation is scope — `prepared_plan` is class-scoped and can't use the function-scoped built-in. A simpler approach is to override the scope in `conftest.py`:

```python
@pytest.fixture(scope="session")
def jira_issue(request):
    """Session-scoped override of pytest-jira's jira_issue fixture."""
    def wrapper(issue_id):
        jira_plugin = request.config.pluginmanager.getplugin("jira_plugin")
        if jira_plugin:
            try:
                return not jira_plugin.is_issue_resolved(issue_id)
            except requests.RequestException:
                return None
        return None
    return wrapper
```

This eliminates the entire `utilities/jira_helpers.py` module and the `jira_issue_open` fixture — `prepared_plan` would just take `jira_issue` directly.

myakove: From the plugin docs, e en simpler



Besides a test marker, you can also use the added jira_issue fixture. This enables examining issue status mid test and not just at the beginning of a test. The fixture return a boolean representing the state of the issue. If the issue isn't found, or the jira plugin isn't loaded, it returns None.

NICE_ANIMALS = ["bird", "cat", "dog"]

def test_stuff(jira_issue):
    animals = ["dog", "cat"]
    for animal in animals:
        if animal == "dog" and jira_issue("ORG-1382") is True:
            print("Issue is still open, cannot check for dogs!")
            continue
        assert animal in NICE_ANIMALS

krcmarik:
> > Honestly I am not really sure If we should workaround what I believe is either bug in forklift or the vmware environment misconfiguration in the test suite. We are supposed to get almost immediate updates from VMware about any inventory changes. If we do not have them within 300s, something is wrong and forcing inventory refresh in the test suite has two consequences:
> > 
> > 1. We are imo covering bugs of the tested product in the test suite which goes against the entire point of test automation.
> > 2. If the problem keeps happening and It seems it does we extend the test run time by 300s with each failure of clone VM's discovery in the forklift inventory.
> 
> You are right, but we have 2 options. Fail the test on the inventory bug or handle it in the test for know till the bug is fixed (the wa taid to Jira ticket) The Jira way it the right way to handle it, while the bug is open we workaround it, once it's closed we go back to the original code flow without the workaround
> 
> I think this option is better for us, we do not want to not test the migration completely because the inventory bug.

That makes sense, if we revert it at some point once the bug is fixed/identified. Just a little heavy PR for a workaround:) But It's a good point that we rather test migrations.

utilities/provider_inventory.py
prabinovRedhat: The two loops serve distinct purposes and cannot be collapsed without losing the retry-only-failures pattern. Loop 1 collects failures across all cloned VMs. Loop 2 retries ONLY the failed subset after a forced provider refresh — retrying everything would be wasteful and mask whether the refresh actually helped. Added a code comment in f79f166c explaining the O(N × inventory_timeout) cumulative timeout design.

utilities/provider_inventory.py
prabinovRedhat: Fixed in f79f166c — changed to f-string as requested.

utilities/copyoffload_datastore.py
prabinovRedhat: Fixed in f79f166c — replaced all 4 hardcoded string literals ('secondary_datastore_id', 'non_xcopy_datastore_id') with the existing constants SYMBOLIC_SECONDARY_DATASTORE and SYMBOLIC_NON_XCOPY_DATASTORE.

libs/forklift_inventory.py
prabinovRedhat: Fixed in f79f166c — added a full docstring to the hosts property per AGENTS.md format.

utilities/jira_helpers.py
prabinovRedhat: Fixed in f79f166c — replaced pytest.skip() with LOGGER.warning(...); return None. The utility now logs a warning and returns None (same as Jira unavailable path) when connection fails with SKIP strategy, never calling pytest.skip() from a non-fixture context.

utilities/provider_inventory.py
prabinovRedhat: Good call. Added a code comment in f79f166c explaining the design: 'Note: total wait is O(N × inventory_timeout) in the worst case — each VM is waited independently because VMs appear in inventory at different times after cloning.'

utilities/provider_inventory.py
prabinovRedhat: Agreed — the workaround is gated by Jira MTV-6072. When that issue is resolved, workaround_active becomes False and the code falls through to the plain wait_for_vm path. No workaround code needs to be deleted — the gate handles it automatically.

libs/providers/vmware.py
prabinovRedhat: Correct analysis — applied the simplification in 80eefaf. Now compares symbolic keys directly instead of resolving to MoID: `if disk_datastore_id == SYMBOLIC_SECONDARY_DATASTORE and secondary_datastore:` → `resolved = secondary_datastore`. The MoID round-trip is gone. Literal MoID values fall through to the else branch and use `get_obj` as before.

AGENTS.md
```diff
@@ -885,6 +885,7 @@ tests_params: dict = {
 | `copyoffload`         | No       | True to enable copy-offload (XCOPY) migration         |
 | `xfs_compatibility`   | No       | True to enable XFS v4 filesystem compatibility        |
 | `migrate_shared_disks`| No       | True to enable shared disk migration at plan level    |
+| `inventory_timeout`   | No       | Max seconds for Forklift inventory waits (default 300)|
```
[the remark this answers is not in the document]
prabinovRedhat: Fixed in 80eefaf — updated to 'Per-VM Forklift inventory wait timeout, in seconds' to clarify it's a per-VM wait, not an overall cap.

utilities/jira_helpers.py
prabinovRedhat: Fixed in 9e6cf31 — the 'plugin not found' warning was misplaced: on the IGNORE strategy path the plugin IS found but the connection error was silently swallowed, causing the misleading log. Restructured: 'not found' warning is now in the else branch (fires only when jira_plugin is None), and a separate warning logs the suppressed error for IGNORE/unrecognised strategy.

prabinovRedhat: Fixed in 80eefaf — added `LOGGER.warning(f"Jira plugin '{JIRA_PLUGIN_NAME}' not found; treating issue '{issue_id}' as unavailable")` before the final `return None` when the plugin is not installed.

utilities/provider_inventory.py
prabinovRedhat: No change — `Ready` is intentional here. The `_refresh` patch triggers a reconciliation loop; `Ready=True` is sufficient for inventory repopulation to start. `Validated=True` checks that the full validation cycle completed, which is not required for this workaround. Using `Ready` also avoids false-positive failures when a provider has pre-existing validation warnings but a healthy inventory. If this reasoning is incorrect please let us know and we will switch to `Validated`.

utilities/jira_helpers.py
prabinovRedhat: Good suggestion. For this PR, keeping the `is_jira_issue_open` utility avoids scope changes that could affect other class-scoped fixtures. The session-scoped fixture approach is tracked as a follow-up refactor once we confirm it works correctly with the pytest-jira plugin internal API.

utilities/copyoffload_datastore.py
prabinovRedhat: Added a defensive comment in 80eefaf — current callers only pass symbolic keys (`SYMBOLIC_SECONDARY_DATASTORE`, `SYMBOLIC_NON_XCOPY_DATASTORE`), but the guard is kept to protect against future callers passing unexpected values.

libs/providers/vmware.py
prabinovRedhat: Fixed in 80eefaf — removed the MoID round-trip. The comparison now uses the symbolic key directly: `if disk_datastore_id == SYMBOLIC_SECONDARY_DATASTORE and secondary_datastore:`. Eliminates the redundant resolve→compare pattern.

libs/forklift_inventory.py
```diff
@@ -406,10 +418,111 @@ def __init__(self, client: DynamicClient, provider_name: str, namespace: str) ->
...
+            f"Waiting for datastores {sorted(requested_ids)} to appear in Forklift inventory "
+            f"for provider '{self.provider_name}'..."
+        )
+
+        def _check_datastores() -> list[dict[str, Any]] | None:
+            storages = self.storages
+            found_ids = _extract_storage_ids(storages)
+            if requested_ids - found_ids:
+                return None
+            return [storage for storage in storages if storage.get("id") in requested_ids]
```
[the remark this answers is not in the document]
prabinovRedhat: Fixed — added docstring to `_check_datastores()` documenting that it returns matching storage entries when all requested IDs are present, or None if any are still missing.

libs/forklift_inventory.py
prabinovRedhat: Fixed in 216bc13 — restructured to `except TimeoutExpiredError: pass` followed by the custom raise after the try/except block. This makes the raise reachable and eliminates the dummy unreachable raise entirely. Applied to both wait_for_hosts and wait_for_datastores.

libs/forklift_inventory.py
```diff
@@ -406,10 +418,116 @@ def __init__(self, client: DynamicClient, provider_name: str, namespace: str) ->
...
+            ):
+                if sample:
+                    LOGGER.info(f"Found {len(sample)} hosts in inventory for provider '{self.provider_name}'")
+                    return sample
+        except TimeoutExpiredError:
+            pass
+
+        raise TimeoutExpiredError(
+            f"No hosts appeared in Forklift inventory for provider '{self.provider_name}' after {timeout}s"
+        )
```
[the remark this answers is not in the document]
prabinovRedhat: TRY003 is not enabled in the project's ruff configuration (pyproject.toml only selects PLC0415). Pre-commit passes without this change. No action needed.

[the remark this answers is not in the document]
prabinovRedhat: Fixed — moved inline TimeoutExpiredError messages to module-level formatter functions _hosts_timeout_error() and _datastores_timeout_error(). TRY003 violations cleared.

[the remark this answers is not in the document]
prabinovRedhat: TRY003 is not enabled in the project ruff configuration (pyproject.toml only selects PLC0415). Pre-commit passes without this change. Skipping.

prabinovRedhat:
## Timeout behavior update

Revised the VM inventory wait logic to reduce total wait time when the MTV-6072 bug is active:

**Previous approach:** Try each VM for the full `inventory_timeout` (300s default) → if any fail, force refresh → retry → worst case: `N × inventory_timeout + refresh_time + N × inventory_timeout`

**New approach (quick-check-then-patch):** Quick check each VM for 30s first → if any fail, force refresh → retry only the failed VMs for the full `inventory_timeout` → worst case: `N × 30s + refresh_time + failed_N × inventory_timeout`

**When MTV-6072 is resolved** (`workaround_active = False`): the patch/refresh block is skipped entirely — runs plain `wait_for_vm(timeout=inventory_timeout)` per VM. Maximum is still `N × inventory_timeout`, but in practice VMs sync in seconds once the Forklift bug is fixed, so actual wait is much shorter.

prabinovRedhat:
It would be great if we could merge this PR soon. There isn't much value in merging it after the bug is fixed, so if possible, I'd appreciate getting it merged as soon as possible since our CI is currently failing because of it.

There's also very little risk in merging it now, once the bug is fixed, the workaround will no longer be used because it's guarded by the Jira marker.

@myakove @tshefi

utilities/jira_helpers.py
prabinovRedhat: Fixed — replaced the custom jira_issue_open conftest fixture and is_jira_issue_open utility with pytest-jira built-in jira_issue fixture. prepared_plan now receives jira_issue directly as a parameter and passes it through to wait_for_cloned_vms_in_forklift_inventory. utilities/jira_helpers.py removed.

prabinovRedhat: Reverted — pytest-jira built-in jira_issue fixture is function-scoped and cannot be used in class-scoped prepared_plan (scope mismatch causes all tests to fail at setup). The session-scoped jira_issue_open wrapper is required. Simplified jira_helpers.py to remove the pytest.skip call (per myakove original comment) while keeping the correct scope.

prabinovRedhat: Fixed — overriding jira_issue with a session-scoped fixture in conftest.py (per krcmarik suggestion). pytest-jira built-in is function-scoped; the override gives it session scope so prepared_plan (class-scoped) can use it without scope mismatch. jira_helpers.py module removed entirely.

